PR #2, opened 2019-02-14
fix README link to .so

gkorland:


gkorland:
@gavrie added Mac too

PR #6, opened 2019-04-27
Breaking into scope blocks to make sure all memory allocated before we

gkorland:
Make sure before we switch to Redis allocator the rest will be out of scope


gavrie:
LGTM

PR #29, opened 2019-09-03
add redis buffer

gkorland:


src/redismodule.rs
```diff
@@ -80,3 +83,43 @@ impl Drop for RedisString {
...
+
+    pub fn to_string(&self) -> Result<String, FromUtf8Error> {
+        unsafe {
+            let slice = slice::from_raw_parts(self.buffer as *const u8, self.len);
+            String::from_utf8(slice.to_vec())
+        }
+    }
+}
+
+impl Index<usize> for RedisBuffer {
```
gavrie: Implementing `Index` here seems a bit overkill to me.
The idiom I usually see is to `impl AsRef` to access the underlying slice, and then use it like this:
`buffer.as_ref()[0]`

To implement this:
```
impl AsRef<[u8]> for RedisBuffer {
    fn as_ref(&self) -> &[u8] {
        unsafe { slice::from_raw_parts(self.buffer as *const u8, self.len) }
    }
}
```

Then, you no longer need to use `slice::from_raw_parts` in multiple methods. Instead, `to_string` becomes:

```
    pub fn to_string(&self) -> Result<String, FromUtf8Error> {
        String::from_utf8(self.as_ref().to_vec())
    }
```

gkorland: done

PR #32, opened 2019-09-03
Use the latest redismodule.h header version

Kerollmops:


gavrie:
@Kerollmops thanks for the contribution. I've updated the header file to use our slightly different format (that we also use in RediSearch). Functionally they should be equivalent.

PR #33, opened 2019-09-04
Introduce the `experimental-api` feature

Kerollmops:
This way it will be possible to enable or disable the Redis experimental-api and therefore not risk using them by mistake.

Every future Rust bindings above experimental function must be marked like so:
```rust
#[cfg(feature = "experimental-api")]
pub fn get_thread_safe_context() -> *mut RedisModuleCtx {
    unsafe { RedisModule_GetThreadSafeContext.unwrap()(std::ptr::null_mut()) }
}
```


gavrie:
Merged.

PR #37, opened 2019-09-22
add missing ThreadSafeContext methods and extend RM_CALL

gkorland:


src/raw.rs
```diff
@@ -158,16 +157,27 @@ pub fn call_reply_integer(reply: *mut RedisModuleCallReply) -> c_longlong {
...
+    unsafe { RedisModule_CallReplyLength.unwrap()(reply) }
+}
+
 pub fn call_reply_string_ptr(reply: *mut RedisModuleCallReply, len: *mut size_t) -> *const c_char {
     unsafe { RedisModule_CallReplyStringPtr.unwrap()(reply, len) }
 }
 
 pub fn call_reply_string(reply: *mut RedisModuleCallReply) -> String {
     unsafe {
-        let len: *mut size_t = null_mut();
```
gavrie: Good find... this was a nasty bug.

src/key.rs
```diff
@@ -66,7 +66,7 @@ impl RedisKey {
         self.key_inner == null_key
     }
 
-    pub fn read(&self) -> Result<Option<String>, Error> {
+    pub fn read(&self) -> Result<Option<String>, RedisError> {
```
gavrie: We need to tidy up our usage of `Error` vs `RedisError`. Need to think about this for a bit.

src/context.rs
```diff
@@ -26,6 +26,22 @@ impl Context {
...
     }
 
+    #[cfg(feature = "experimental-api")]
+    pub fn get_thread_safe_context() -> Self {
+        let ctx = unsafe { raw::RedisModule_GetThreadSafeContext.unwrap()(ptr::null_mut()) };
+        Context::new(ctx)
+    }
+
+    #[cfg(feature = "experimental-api")]
+    pub fn lock(&self) {
```
gavrie: It would be nice to model this like `std::sync::Mutex`:
- `get_thread_safe_context` would return a guard object, instead of the `Context` itself.
- Calling `lock` in the guard object would return the `Context` stored inside the guard.
- `impl Drop` on the guard would unlock the `Context`.

PR #58, opened 2019-12-24
Add support for hash_get/hash_set and Sring operations

gkorland:


src/key.rs
```diff
@@ -74,6 +74,15 @@ impl RedisKey {
         };
         Ok(val)
     }
+
+    pub fn hash_get(&self, field: &str) -> Result<Option<RedisString>, RedisError> {
+        let val = if self.is_null() {
```
gavrie: Isn't this unnecessary since `hash_get_key` already does null detection? Or is it meant to improve performance by avoiding the need to open the key?

gkorland: good question, I followed the same pattern you did in line 70 with read()

PR #62, opened 2020-02-23
cleaner way to use redis allocator

siscia:
This needs to be formatted, by for some exoteric reason today my formatter does not work on this machine.

Anyway, this PR allow to use the redis allocator without the ugly check at compile time. The tradeoff is limiting the name of the redis module, now it is limited to 64-1 chars, of course it can be increased.

I prepare this PR in response to this tweet: https://twitter.com/croloris/status/1231318432068653056 and the related post: https://engineering.redislabs.com/posts/using-the-redis-allocator-in-rust/

I will complete and format the PR later :)

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -79,19 +79,24 @@ macro_rules! redis_module {
             _argv: *mut *mut raw::RedisModuleString,
             _argc: c_int,
         ) -> c_int {
-            {
+            let mut name_buffer = [0; 64];
+                  let mut dest = name_buffer.as_mut_ptr();
+            for byte in $module_name.chars() {
+                unsafe {
+                    *dest = byte as i8;
+                    dest = dest.add(1);
+                }
+            }
                 // We use an explicit block here to make sure all memory allocated before we
                 // switch to the Redis allocator will be out of scope and thus deallocated.
-                let module_name = CString::new($module_name).unwrap();
                 let module_version = $module_version as c_int;
 
                 if unsafe { raw::Export_RedisModule_Init(
                     ctx,
-                    module_name.as_ptr(),
+                    name_buffer.as_ptr() as *const std::os::raw::c_char,
                     module_version,
                     raw::REDISMODULE_APIVER_1 as c_int,
                 ) } == raw::Status::Err as c_int { return raw::Status::Err as c_int; }
-            }
 
             if true {
                 redis_module::alloc::use_redis_alloc();
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -65,7 +65,7 @@ macro_rules! redis_module {
               ]),* $(,)*
         ] $(,)*
     ) => {
-        use std::os::raw::c_int;
+        use std::os::raw::{c_int, c_char};
         use std::ffi::CString;
         use std::slice;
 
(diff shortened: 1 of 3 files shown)


siscia:
`cargo test --example hello` still fails, but I believe you guys are in a better position to fix it.

The formatting was not broken, was only inside a macro and it was not suppose to work in first place :)

I would also suggest to run `clippy` on the project, there are several low hanging fruit that we could rip (mostly `String` -> `&str`), moreover there are some warnings that are worth to investigate.

It will be kinda noise because a lot of time we deference raw pointers without unsafe, I fix it wrapping the raw pointers in a rust structure.

Also, a good idea would be to add an integration test to the CI, I got a reasonably script here: https://github.com/RedBeardLab/rediSQL/blob/master/.github/workflows/rust.yml

It is a little more complex of what you guys need (it use multiple redis version and multiple test for different version of RediSQL) but it is quite fast ~5min to run and it could be made faster if that is a problem.

Cheers,
Simone

src/macros.rs
```diff
@@ -79,25 +79,24 @@ macro_rules! redis_module {
...
-                    raw::REDISMODULE_APIVER_1 as c_int,
-                ) } == raw::Status::Err as c_int { return raw::Status::Err as c_int; }
+            let mut name_buffer = [0; 64];
+            let mut dest = name_buffer.as_mut_ptr();
+            for byte in $module_name.chars() {
+                unsafe {
+                    *dest = byte as i8;
+                    dest = dest.add(1);
+                }
             }
```
gavrie: I'd prefer to do this as follows (there is no need to distinguish between UTF-8 and plain bytes here, so no need to call `chars()`:
```
        std::ptr::copy(
            module_name.as_ptr(),
            name_buffer.as_mut_ptr(),
            module_name.len(),
        );
```

gavrie:
@siscia Thanks for your PR! I very much like the concept.

Do you see any reason not to use `std::ptr::copy` as I suggested in my comment?
If not, I'll update the PR and merge.

siscia:
@gavrie actually it seems a better idea to use `std::ptr::copy` it should compile easily as well.
I don't see reason for not using it.

Before to merge I would also consider to fix `cargo test --example hello` :)

gkorland:
>Before to merge I would also consider to fix cargo test --example hello :)
What is broken?

siscia:
On this PR.

```
:~/redismodule-rs$ cargo test --example hello
   Compiling redis-module v0.7.0 (/home/smosciat/redismodule-rs)
    Finished test [unoptimized + debuginfo] target(s) in 6.38s
     Running target/debug/examples/hello-8fe32d15b93b44ac
thread panicked while processing panic. aborting.
error: test failed, to rerun pass '--example hello

Caused by:
  process didn't exit successfully: `/home/smosciat/redismodule-rs/target/debug/examples/hello-8fe32d15b93b44ac` (signal: 4, SIGILL: illegal instruction)
```

I believe the problem is that cargo `test` and the `--example` flag don't work very well together.

In my understading cargo does not understand that it is a test and it try to allocate memory through redis, which of course fail.

gavrie:
> I believe the problem is that cargo `test` and the `--example` flag don't work very well together.
> 
> In my understading cargo does not understand that it is a test and it try to allocate memory through redis, which of course fail.

@siscia I looked into this a bit, and your theory seems to be correct:
https://github.com/rust-lang/rust/issues/59168#issuecomment-472647807

I'm adding a workaround for this that was suggested in several places: Create a Cargo `feature` named `test` and check on that in the code.

gavrie:
@siscia I have merged your commits into #68, and will merge from there to master after a bit more testing. Thanks!

PR #63, opened 2020-03-05
Add Support for CreateCommand Keys

grippy:
The CreateCommand firstkey, lastkey, and keystep defaults aren't working for a module I'm trying to work on. 

I'd really like to make this change non-breaking. But my macro fu is noob-level (at best) and I don't know how to glob all the values after the `$flags:expr` (or even if this is something rust macros support?) and then pass these values as a vec into `redis_command!`. This way we could set defaults instead of passing `1, 1, 1` for most commands.

### DIFF
--- a/examples/data_type.rs
+++ b/examples/data_type.rs
@@ -83,7 +83,7 @@ redis_module! {
         MY_REDIS_TYPE,
     ],
     commands: [
-        ["alloc.set", alloc_set, "write"],
-        ["alloc.get", alloc_get, "readonly"],
+        ["alloc.set", alloc_set, "write", 1, 1, 1],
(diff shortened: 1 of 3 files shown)


gavrie:
@grippy Thanks for your PR! 

As far as I can tell, `macro_rules` has no way to allow making this optional (or at least, I couldn't find a way to do it).

I have not problem with requiring the 3 extra arguments. It breaks backwards compatibility, but this module's API is not really stable yet so other modules will need to adjust.

We'll make this a new version to make that clear.

gavrie:
@gkorland WDYT on making the incompatible change (see my previous comment)?

gkorland:
@gavrie we can live with this breaking change

PR #66, opened 2020-04-01
SubscribeKeyspaceEvent & Timer API's

grippy:
This PR adds support for these API methods:

- `RedisModule_CreateTimer`
- `RedisModule_StopTimer`
- `RedisModule_GetTimerInfo`
- `RedisModule_SubscribeToKeyspaceEvents`

A few things I want to point out:

1. This was my first attempt at doing anything w/ Rust FFI & C Bindings. I really struggled with figuring out how to read data from a `*mut *mut c_void` without destroying the original pointer. If there's a better way to do this please let me know.

2. I'm currently passing `&str` for the `data` arg in `create_timer` only to convert it to a CString. Since the arg parser returns `Vec<String>` it might make more sense to change the `data` arg to String. What do you think?

3. I'm not entirely sure what the cleanest API here is for returning the "data" stored with a timer when you call `RedisModule_GetTimerInfo` or `RedisModule_StopTimer`. What I have here feels a little clunky returning a tuple. Maybe it needs a new return type or Enum?

I haven't created an issue to track this. Let me know if you think I should.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -40,6 +40,30 @@ impl Context {
         unsafe { raw::RedisModule_ThreadSafeContextUnlock.unwrap()(self.ctx) };
     }
 
+    #[cfg(feature = "experimental-api")]
+    pub fn create_timer(
+        &self,
+        period: u64,
+        callback: raw::RedisModuleTimerProc,
+        data: &str,
+    ) -> u64 {
+        raw::create_timer(self.ctx, period, callback, data)
+    }
+
+    #[cfg(feature = "experimental-api")]
+    pub fn stop_timer(&self, id: u64) -> i32 {
+        raw::stop_timer(self.ctx, id)
+    }
+
+    #[cfg(feature = "experimental-api")]
+    pub fn subscribe_to_keyspace_events(
+        &self,
+        types: i32,
+        callback: raw::RedisModuleNotificationFunc,
+    ) -> i32 {
+        raw::subscribe_to_keyspace_events(self.ctx, types, callback)
+    }
+
     pub fn log(&self, level: LogLevel, message: &str) {
         let level = CString::new(format!("{:?}", level).to_lowercase()).unwrap();
         let fmt = CString::new(message).unwrap();
--- a/src/context.rs
+++ b/src/context.rs
@@ -50,11 +50,35 @@ impl Context {
         raw::create_timer(self.ctx, period, callback, data)
     }
 
+    // stop_timer_data takes an timer_id and
+    // returns ok = 0, err = 1
     #[cfg(feature = "experimental-api")]
     pub fn stop_timer(&self, id: u64) -> i32 {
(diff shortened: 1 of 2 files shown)


gavrie:
@grippy thanks for your PR!

Prompted by your question on StackOverflow, I worked on my own implementation of this before seeing your PR 😄. It's nice to see your implementation and compare what we both came up with.

My PR is not ready yet, but you can take a look: #67.

> 1. This was my first attempt at doing anything w/ Rust FFI & C Bindings. I really struggled with figuring out how to read data from a `*mut *mut c_void` without destroying the original pointer. If there's a better way to do this please let me know.

Yes, this is a complicated issue. It took me some time to wrap my head around that as well the first time I encountered this. The simplest way would be something like this:

```
let mut remaining: u64 = 0;
let mut data: *mut c_void = std::ptr::null_mut();
                                                                                         
let status: raw::Status = unsafe {
    raw::RedisModule_GetTimerInfo.unwrap()(self.ctx, timer_id, &mut remaining, &mut data)
}
```

> 2. I'm currently passing `&str` for the `data` arg in `create_timer` only to convert it to a CString. Since the arg parser returns `Vec<String>` it might make more sense to change the `data` arg to String. What do you think?

I'm more in favor of keeping the `data` generic, so that the user can pass their preferred type. This is what I did in my implementation.

> 3. I'm not entirely sure what the cleanest API here is for returning the "data" stored with a timer when you call `RedisModule_GetTimerInfo` or `RedisModule_StopTimer`. What I have here feels a little clunky returning a tuple. Maybe it needs a new return type or Enum?

I prefer returning a standard Rust `Result` instead of the raw Redis status, which removes the need for a tuple in most cases. For `GetTimerInfo` specifically, a 2-tuple is fine.

grippy:
@gavrie Thanks for the feedback. I took at a look at your PR and its much cleaner. I tried the example out and it works. I also really like how you're using custom types. This PR no longer makes sense. I think I'm going to close this one in favor of your work. Once you PR lands, I can follow your lead and and see if I can figure out the `RedisModule_SubscribeToKeyspaceEvents` part later.

grippy:
Superseded by:  https://github.com/RedisLabsModules/redismodule-rs/pull/67

gavrie:
> @gavrie Thanks for the feedback. I took at a look at your PR and its much cleaner. I tried the example out and it works. I also really like how you're using custom types. This PR no longer makes sense. I think I'm going to close this one in favor of your work. Once you PR lands, I can follow your lead and and see if I can figure out the `RedisModule_SubscribeToKeyspaceEvents` part later.

@grippy I'm glad you see it this way! Thanks to you I had a chance to delve into this, and both learn a bunch and also implement a useful feature.

I'll be glad to get your PR for `RedisModule_SubscribeToKeyspaceEvents`.

gavrie:
I've merged #67 so you should be all set.

PR #72, opened 2020-04-19
Update bindgen to 0.53

messense:
Closes #59


gavrie:
Thanks @messense ! It took me a while to get to this and understand what was broken.

PR #75, opened 2020-04-27
Add high level wrapper for RedisModule_EmitAOF

### LINKED ISSUE #74: Add high level wrapper for RedisModule_EmitAOF

messense:
Closes #74

### DIFF
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -209,4 +209,9 @@ impl Context {
     pub fn create_string(&self, s: &str) -> RedisString {
         RedisString::create(self.ctx, s)
     }
+
+    #[cfg(feature = "experimental-api")]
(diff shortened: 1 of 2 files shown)


src/context/mod.rs
```diff
@@ -209,4 +209,9 @@ impl Context {
     pub fn create_string(&self, s: &str) -> RedisString {
         RedisString::create(self.ctx, s)
     }
+
+    #[cfg(feature = "experimental-api")]
+    pub fn emit_aof(&self, io: *mut raw::RedisModuleIO, command: &str, args: &[&str]) {
```
gavrie, line 214: The public API of `Context` should not use raw pointers like `*mut raw::RedisModuleIO`; its whole purpose it to provide a safe abstraction over them.

I haven't dived deeply into this, but it seems the correct thing here is to create a wrapper struct around the raw type and pass safe Rust references to it (like `Context` itself).

gavrie (changes requested):

PR #76, opened 2020-05-14
RedisModule_SubscribeToKeyspaceEvents

grippy:
The PR adds support for `RedisModule_SubscribeToKeyspaceEvents`. I would've liked it if the end user could pass Rust callbacks. But this api call doesn't store `data` that we can lookup in the callback like `CreateTimer` does. I also added simple example and some tests for parsing the various notification event abbreviations.


gavrie:
It seems like #78 has a lot of overlap with this. It would be great if you guys could come up with a best-of-breed PR together!

grippy:
Closing this in favor of https://github.com/RedisLabsModules/redismodule-rs/pull/78

Issue #77, opened 2020-05-16
Expose SubscribeToKeyspaceEvents


matobet:
It might be useful to expose in a rust-native way registration of event handlers to keyspace events (using the `RedisModule_SubscribeToKeyspaceEvents` API). E.g. a macro or such.

matobet:
@gavrie could you please review https://github.com/RedisLabsModules/redismodule-rs/pull/78? it's been waiting for review for over a month.

gavrie:
Fixed by #78

PR #78, opened 2020-05-17
Expose SubscribeToKeyspaceEvents (#77)

matobet:
Add optional `event_handlers` field to the `redis_module!` macro that enables
user to specify auto-registered keyspace event handlers.

### DIFF
--- a/examples/events.rs
+++ b/examples/events.rs
@@ -1,0 +1,34 @@
+#[macro_use]
+extern crate redis_module;
+
+use redis_module::{Context, NotifyEvent};
+
+fn on_event(ctx: &Context, event_type: NotifyEvent, event: &str, key: &str) {
+    let msg = format!(
+        "Received event: {:?} on key: {} via event: {}",
+        event_type, key, event
+    );
+    ctx.log_debug(msg.as_str());
+}
+
+fn on_stream(ctx: &Context, _event_type: NotifyEvent, _event: &str, _key: &str) {
+    ctx.log_debug("Stream event received!");
+}
+
+//////////////////////////////////////////////////////
+
+redis_module! {
+    name: "events",
+    version: 1,
+    data_types: [],
+    commands: [],
+    event_handlers: [
+        [@EXPIRED @EVICTED: on_event],
+        [@STREAM: on_stream],
+    ]
+}
+
+//////////////////////////////////////////////////////
+
+#[cfg(test)]
+mod tests {}
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -55,6 +55,44 @@ macro_rules! redis_command {
     }};
 }
 
+#[cfg(feature = "experimental-api")]
+#[macro_export]
+macro_rules! redis_event_handler {
+    (
+        $ctx: expr,
+        $event_type: expr,
+        $event_handler: expr
+    ) => {{
+        extern "C" fn handle_event(
+            ctx: *mut $crate::raw::RedisModuleCtx,
+            event_type: c_int,
+            event: *const c_char,
+            key: *mut $crate::raw::RedisModuleString
+        ) -> c_int {
+            let context = $crate::Context::new(ctx);
+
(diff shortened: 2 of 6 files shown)


examples/events.rs
```diff
@@ -0,0 +1,27 @@
...
+
+//////////////////////////////////////////////////////
+
+redis_module! {
+    name: "events",
+    version: 1,
+    data_types: [],
+    commands: [],
+    event_handlers: [
+        [raw::NotifyEvent::EXPIRED | raw::NotifyEvent::EVICTED, on_event]
```
matobet, line 19: The only thing I dislike about this is the fact that the `raw` module is leaking to the highest-level macro definitions. As an alternative, we could have a "higher-level" enum mapped to this defined in `redismodule.rs`, but for the PoC I decided to go with the minimal approach.

gavrie:
@matobet I like the idea!
I also agree to your comment regarding `raw` and the syntax.
If you want to polish it a bit more that would be great.

gavrie:
It seems like #76 has a lot of overlap with this. It would be great if you guys could come up with a best-of-breed PR together!

matobet:
@gavrie @grippy Thanks for the vote of confidence! I took a stab at a more higher-level macro definitions using the `@TYPE` syntax (see example). I'm still pretty new to the Rust world so feedback is very welcome. :+1:

matobet:
So @gavrie @gkorland do you have any comments/suggestions or do you think we can move forward with merging this as is?

grippy:
@gavrie @gkorland Any comments on this PR? Would love to see this merged if you agree.

gavrie:
> @gavrie @gkorland Any comments on this PR? Would love to see this merged if you agree.

@grippy We'll look at it. Sorry @matobet , somehow missed your updates on this.

src/macros.rs
```diff
@@ -53,6 +53,44 @@ macro_rules! redis_command {
...
+    ) => {{
+        extern "C" fn handle_event(
+            ctx: *mut raw::RedisModuleCtx,
+            event_type: c_int,
+            event: *const c_char,
+            key: *mut raw::RedisModuleString
+        ) -> c_int {
+            let context = Context::new(ctx);
+
+            let redis_key = RedisString::from_ptr(key).unwrap();
```
slavak, line 72: This (and the `event_str.to_str().unwrap()` below) will cause a panic if the strings are not valid UTF-8, which IIRC Redis doesn't guarantee. I believe this would hardly be the only place in the crate where this issue may happen. @gavrie @gkorland Do we have any policy about this? (Asking in general, not for this PR specifically.)

slavak:
I like the macro `@<EVENT_TYPE>` approach! Very ergonomic.

Not necessarily thrilled about the Rust event handlers needing to handle a type from the `raw` module (`NotifyEvent`). The type itself is probably high-level enough to be used, so maybe just re-export that specific one from a higher-level module of the crate?

matobet:
@slavak Thanks! Re-exported `NotifyEvent` from the top-level mod & added the `$crate` qualification to macros.

src/macros.rs
gavrie: @slavak This is indeed a problematic assumption, and I think we make it in several places as you said. Not a problem specific to this PR, we should think about what we want to do here in general.

gavrie:
> I like the macro `@<EVENT_TYPE>` approach! Very ergonomic.

I agree 👍

slavak:
Thanks @matobet!

I think the PR looks good. Will probably take Gavrie a few days to be able to go over it. Apologies.

gavrie:
Finally got to review this.
I looks good! I have a few suggested cleanups: https://github.com/matobet/redismodule-rs/pull/1
@matobet provide you agree with those, please update your PR with my changes and I'll merge upstream.

matobet:
@gavrie Merged, thanks!

gavrie (approved):

PR #80, opened 2020-05-21
add Aux events

gkorland:


src/raw.rs
```diff
@@ -68,6 +68,12 @@ impl From<c_int> for ReplyType {
     }
 }
 
+#[derive(Primitive, Debug, PartialEq)]
+pub enum Aux {
+    Before = REDISMODULE_AUX_BEFORE_RDB as isize,
+    After = REDISMODULE_AUX_AFTER_RDB as isize,
+}
+
```
gavrie, line 76: Is this in preparation for some other future? Where is it used?

gkorland: https://github.com/RedisJSON/RedisJSON/pull/189

PR #87, opened 2020-06-15
Add support for server events (redis 6)

yoav-steinberg:
Should I also add a usage example?
Note that this code was verified for usage when handling https://github.com/RedisJSON/RedisJSON/issues/180, although I haven't pushed the PR for that yet.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -12,6 +12,17 @@ impl ParseCallbacks for RedisModuleCallback {
     fn int_macro(&self, name: &str, _value: i64) -> Option<IntKind> {
         if name.starts_with("REDISMODULE_SUBEVENT_") || name.starts_with("REDISMODULE_EVENT_") {
             Some(IntKind::U64)
+        } else if name.starts_with("REDISMODULE_REPLY_")
+            || name.starts_with("REDISMODULE_KEYTYPE_")
+            || name.starts_with("REDISMODULE_AUX_")
+            || name == "REDISMODULE_OK"
+            || name == "REDISMODULE_ERR"
+        {
+            // These values are used as `enum` discriminants, and thus must be `isize`.
+            Some(IntKind::Custom {
+                name: "isize",
+                is_signed: true,
+            })
         } else {
             None
         }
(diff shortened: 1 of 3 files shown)


src/raw.rs
```diff
@@ -363,3 +363,11 @@ pub fn string_append_buffer(
...
     }
 }
+
+pub fn subscribe_to_server_event(
+    ctx: *mut RedisModuleCtx,
+    event: RedisModuleEvent,
+    callback: RedisModuleEventCallback,
+) -> Status {
+    unsafe { RedisModule_SubscribeToServerEvent.unwrap()(ctx, event, callback).into() }
+}
```
gavrie, line 373: I suggest adding a higher level wrapper above this layer.
Have you seen the code in `src/context/timer.rs`? It can be done in the same way.

build.rs
```diff
@@ -1,9 +1,23 @@
 extern crate bindgen;
 extern crate cc;
 
+use bindgen::callbacks::{IntKind, ParseCallbacks};
 use std::env;
 use std::path::PathBuf;
 
+#[derive(Debug)]
+struct RedisModuleCallback();
```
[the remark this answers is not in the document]
yoav-steinberg: Right, these empty structs keep confusing me. I wonder what in practice does it mean to have a boxed empty struct (memory wise)?

src/raw.rs
yoav-steinberg: The thing here is that `RedisModuleEventCallback` receives a pointer to a struct with extra data about the event where the struct type changes based on which event we chose . So Wapping this nicely means we need to be aware of the types of events that exist and then mapping the `data` type to the correct struct type when when calling the high level callback.
I'm willing to attempt to do this but it means we're more tightly bount between our rust code and redisumode.h, when new events are added it won't work out of the box, we'll need to update our translation layer.
Let me know if I was clear here.

yoav-steinberg: Another issue is that there's no user data here, so there's no place to store the high level callback.

gavrie: Ok, these are all good points. I mainly wanted to bring these up, not sure they're necessary.
Let's do this on an as-needed basis and see where it leads; for now I see to good reason to invest in this.

PR #93, opened 2020-07-09
Changed parse_call_reply to return a BulkString.

MikkelSnitker:
This PR changes the parse_call_reply function to return a BulkString instead of af  SimpleString. 

The current implementation treats all string values, returned by calling a redis function within a module, as a SimpleString. This issue arises if you pass the the return value of call from your module.

The change i propose will treat string values returned from call as a BulkString as BulkStrings can contain line breaks, as the underlying string can contain line brakes

### DIFF
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -120,7 +120,7 @@ impl Context {
                 Ok(RedisValue::Array(vec))
             }
             raw::ReplyType::Integer => Ok(RedisValue::Integer(raw::call_reply_integer(reply))),
-            raw::ReplyType::String => Ok(RedisValue::SimpleString(raw::call_reply_string(reply))),
+            raw::ReplyType::String => Ok(RedisValue::BulkString(raw::call_reply_string(reply))),
             raw::ReplyType::Null => Ok(RedisValue::Null),
         }
     }


yoav-steinberg:
Another option would be to remove the distinction between Bulk/Simple strings from the `RedisValue` enum. This might be more in line with how the module API works (there's no such distinction).
Then all we'll need to do is change the implementation in `Context.reply()` so it decides how to handle the string values based on how many lines there are in it. I think this is what the original module design was aiming for.

phartleyp:
I was about to re-submit this exact change before noticing that it had previously been submitted.

I spent way too long troubleshooting code that was affected by this issue.  For example, this code looks harmless but contains a bug:
```rust
/// Redis module command that returns the value of at key `key`
pub fn cmd_example(ctx: &Context, _args: Vec<RedisString>) -> RedisResult {
    ctx.call("GET", &["key"])
}
```

This code works fine if the key's value does not contain `\r\n` (i.e. it is a SimpleString).  However, if the key's value contains `\r\n` (i.e. it should be a BulkString), redismodule-rs still maps it in to a SimpleString.  This causes the Redis client (redis-py in my case) to hang because RESP protocol is not respected.

It seems that this pull request is delayed because there is desire to make `Context::call()` more intelligent about the RedisValue enum that it returns.  But until that happens, using `BulkString` instead of `SimpleString` would still be a safer choice.

PR #94, opened 2020-07-23
add Context::free_thread_safe_context()

zhao-lang:
This PR creates a wrapper for `RedisModule_FreeThreadSafeContext` to the experimental API.

Without calling `RedisModule_FreeThreadSafeContext`, the thread safe contexts never seem to be freed.

### DIFF
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -34,6 +34,11 @@ impl Context {
     }
 
     #[cfg(feature = "experimental-api")]
+    pub fn free_thread_safe_context(&self) {
+        unsafe { raw::RedisModule_FreeThreadSafeContext.unwrap()(self.ctx) };
+    }
+
+    #[cfg(feature = "experimental-api")]
     pub fn lock(&self) {
         unsafe { raw::RedisModule_ThreadSafeContextLock.unwrap()(self.ctx) };
     }


src/context/mod.rs
```diff
@@ -33,6 +33,11 @@ impl Context {
         Context::new(ctx)
     }
 
+    #[cfg(feature = "experimental-api")]
+    pub fn free_thread_safe_context(&self) {
```
gkorland, line 37: I would have expect that after that command the context ownership will be "lost" and the context will be freed. @gavrie ?

zhao-lang: That does feel more idiomatic. `Context` doesn't to currently implement custom `Drop`. Also looks like `Context` would need a field to distinguish between regular and thread safe contexts as `RedisModule_FreeThreadSafeContext` should only be called on the thread safe versions?

gavrie: @zhao-lang @gkorland I've written a new PR #95 which solves this idiomatically by using `Drop` to handle the freeing, and `Deref` to handle the locking.

gavrie: @zhao-lang if you like the way this is done in #95, please close your PR and I'll merge that one.

zhao-lang:
closing in favor of #95

gavrie:
> closing in favor of #95

@zhao-lang I've merged #95. Let me know if it works for you!

zhao-lang:
> > closing in favor of #95
> 
> @zhao-lang I've merged #95. Let me know if it works for you!

Just tested it out, works as expected!

PR #98, opened 2020-09-27
Add method for converting RedisString to &str

slavak:
This makes it more ergonomic to work with `RedisString` objects.

### DIFF
--- a/src/redismodule.rs
+++ b/src/redismodule.rs
@@ -3,6 +3,7 @@ use std::os::raw::{c_char, c_void};
 use std::slice;
 use std::str;
 use std::string::FromUtf8Error;
+use std::str::Utf8Error;
 
 pub use crate::raw;
 pub use crate::rediserror::RedisError;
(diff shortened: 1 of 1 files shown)


src/redismodule.rs
```diff
@@ -7,6 +7,7 @@ use std::string::FromUtf8Error;
 pub use crate::raw;
 pub use crate::rediserror::RedisError;
 pub use crate::redisvalue::RedisValue;
+use bitflags::_core::str::Utf8Error;
```
gavrie, line 10: 1. The import is wrong, should be `use std::str::Utf8Error`.
2. Also, please put it higher up together with other `std` imports.
3. Finally, as long as you've imported it, please change the `from_ptr` return type accordingly.

gavrie (changes requested):

slavak:
Well that's an embarrassing mistake...

Fixed all 3 points.

gavrie (dismissed):

gavrie (approved):

PR #99, opened 2020-09-28
Fix usage of RedisModule_HashGet in raw::hash_get

slavak:
The Rust value "0" was being used as the terminating null pointer signalling the end of the variadic arguments. Unfortunately the type of this expression is i32, which creates a "read-past-the-end" condition in the calling C code. For calls to RedisModule_HashGet with a single argument this seems to usually work correctly, but trying to do the with 2 or more fields crashes with a segmentation fault.

### DIFF
--- a/src/raw.rs
+++ b/src/raw.rs
@@ -240,7 +240,7 @@ pub fn hash_get(key: *mut RedisModuleKey, field: &str) -> *mut RedisModuleString
             REDISMODULE_HASH_CFIELDS as i32,
             CString::new(field).unwrap().as_ptr(),
             &res,
-            0,
+            ptr::null::<c_char>(),
         );
     }
     res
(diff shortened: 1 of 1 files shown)


src/raw.rs
```diff
@@ -240,7 +240,7 @@ pub fn hash_get(key: *mut RedisModuleKey, field: &str) -> *mut RedisModuleString
             REDISMODULE_HASH_CFIELDS as i32,
             CString::new(field).unwrap().as_ptr(),
             &res,
-            0,
+            ptr::null::<c_char>(),
```
gavrie, line 243: Why `c_char`? Should be `ptr::null::<RedisModuleString>()`, right?

Also: The same fix should be made to `HashSet` below.

gavrie (changes requested):

src/raw.rs
slavak: Well technically since we're passing in `REDISMODULE_HASH_CFIELD` the pointers are to C-strings. We could also make this a `ptr::null::<core::ffi::c_void>()`.

slavak: I've pushed a rebase that applies the fix to `hash_set` as well.

gavrie: Ouch... this API is even uglier than I thought. You're right.

gavrie (approved):

gavrie (approved):

PR #105, opened 2020-10-03
Add infrastructure for exposing hash-multi-get functionality

slavak:
This commit adds a `hash_get_multi` function in the `raw` module, and a private
`hash_mget_key` function in the `key` module. In the interests of avoiding code
duplication, the corresponding single-get functions are removed, and the public
RedisKey*.hash_get functionality is implemented via multi-get.

Public-facing RedisKey* methods for hash multi-get should be the subject of their
own commit.

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -85,7 +85,7 @@ impl RedisKey {
         let val = if self.is_null() {
             None
         } else {
-            hash_get_key(self.ctx, self.key_inner, field)
+            hash_mget_key(self.ctx, self.key_inner, &[field])?.pop().unwrap()
         };
         Ok(val)
     }
@@ -150,7 +150,7 @@ impl RedisKeyWritable {
     }
 
     pub fn hash_get(&self, field: &str) -> Result<Option<RedisString>, RedisError> {
-        Ok(hash_get_key(self.ctx, self.key_inner, field))
+        Ok(hash_mget_key(self.ctx, self.key_inner, &[field])?.pop().unwrap())
     }
 
     pub fn set_expire(&self, expire: Duration) -> RedisResult {
@@ -246,17 +246,34 @@ fn read_key(key: *mut raw::RedisModuleKey) -> Result<String, Utf8Error> {
     )
 }
 
-fn hash_get_key(
-    ctx: *mut raw::RedisModuleCtx,
-    key: *mut raw::RedisModuleKey,
-    field: &str,
-) -> Option<RedisString> {
-    let res = raw::hash_get(key, field);
-    if res.is_null() {
-        None
-    } else {
-        Some(RedisString::new(ctx, res))
+/// Get an arbitrary number of hash fields from a key by batching calls
+/// to `raw::hash_get_multi`.
+pub fn hash_mget_key<T>(ctx: *mut raw::RedisModuleCtx,
+                        key: *mut raw::RedisModuleKey,
+                        fields: &[T])
+    -> Result<Vec<Option<RedisString>>, RedisError>
+    where T: Into<Vec<u8>> + Clone
+{
+    use crate::redisraw::bindings::RedisModuleString;
+    const BATCH_SIZE: usize = 12;
+
+    let mut values: Vec<Option<RedisString>> = Vec::with_capacity(fields.len());
+    let mut values_raw: [*mut RedisModuleString; BATCH_SIZE] =
+        [std::ptr::null_mut(); BATCH_SIZE];
+
+    for chunk_fields in fields.chunks(BATCH_SIZE) {
+        let mut chunk_values = &mut values_raw[..chunk_fields.len()];
+        raw::hash_get_multi(key, chunk_fields, &mut chunk_values)?;
+        values.extend(chunk_values.iter().map(|ptr|
+            if ptr.is_null() {
+                None
+            } else {
+                Some(RedisString::new(ctx, *ptr))
+            }
+        ));
     }
+
+    Ok(values)
 }
 
 fn to_raw_mode(mode: KeyMode) -> raw::KeyMode {
--- a/src/key.rs
+++ b/src/key.rs
@@ -85,7 +85,9 @@ impl RedisKey {
         let val = if self.is_null() {
             None
         } else {
-            hash_mget_key(self.ctx, self.key_inner, &[field])?.pop().unwrap()
+            hash_mget_key(self.ctx, self.key_inner, &[field])?
+                .pop()
+                .expect("hash_mget_key should return vector of same length as input")
         };
         Ok(val)
     }
@@ -150,7 +152,9 @@ impl RedisKeyWritable {
     }
 
     pub fn hash_get(&self, field: &str) -> Result<Option<RedisString>, RedisError> {
-        Ok(hash_mget_key(self.ctx, self.key_inner, &[field])?.pop().unwrap())
+        Ok(hash_mget_key(self.ctx, self.key_inner, &[field])?
+            .pop()
+            .expect("hash_mget_key should return vector of same length as input"))
     }
 
     pub fn set_expire(&self, expire: Duration) -> RedisResult {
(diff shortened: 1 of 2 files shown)


src/key.rs
```diff
@@ -246,17 +246,34 @@ fn read_key(key: *mut raw::RedisModuleKey) -> Result<String, Utf8Error> {
...
-        None
-    } else {
-        Some(RedisString::new(ctx, res))
+/// Get an arbitrary number of hash fields from a key by batching calls
+/// to `raw::hash_get_multi`.
+pub fn hash_mget_key<T>(ctx: *mut raw::RedisModuleCtx,
+                        key: *mut raw::RedisModuleKey,
+                        fields: &[T])
+    -> Result<Vec<Option<RedisString>>, RedisError>
+    where T: Into<Vec<u8>> + Clone
```
gavrie, line 255: I'm planning to look a bit deeper to understand the allocations and see if there are inefficiencies that can be avoided.

src/raw.rs
```diff
@@ -232,18 +232,59 @@ pub fn string_dma(key: *mut RedisModuleKey, len: *mut size_t, mode: KeyMode) ->
...
+                                         ptr::null::<c_char>())
+        }};
+        ($($args:expr),*) => { unsafe {
+            RedisModule_HashGet.unwrap()(
+                key, REDISMODULE_HASH_CFIELDS as i32,
+                $($args),*,
+                ptr::null::<c_char>()
+            )
+        }};
+    }
```
gavrie, line 257: Very nice, exactly what I was aiming for!

src/raw.rs
```diff
@@ -232,18 +232,59 @@ pub fn string_dma(key: *mut RedisModuleKey, len: *mut size_t, mode: KeyMode) ->
...
+        1  => RM!(f!(), v!()),
+        2  => RM!(f!(), v!(), f!(), v!()),
+        3  => RM!(f!(), v!(), f!(), v!(), f!(), v!()),
+        4  => RM!(f!(), v!(), f!(), v!(), f!(), v!(), f!(), v!()),
+        5  => RM!(f!(), v!(), f!(), v!(), f!(), v!(), f!(), v!(), f!(), v!()),
+        6  => RM!(f!(), v!(), f!(), v!(), f!(), v!(), f!(), v!(), f!(), v!(), f!(), v!()),
+        7  => RM!(f!(), v!(), f!(), v!(), f!(), v!(), f!(), v!(), f!(), v!(), f!(), v!(),
+                  f!(), v!()),
+        8  => RM!(f!(), v!(), f!(), v!(), f!(), v!(), f!(), v!(), f!(), v!(), f!(), v!(),
+                  f!(), v!(), f!(), v!()),
```
gavrie, line 272: I wonder whether we really need 12 or can stop at 8 or so. Sounds a bit excessive (but not important enough to obsess on).

src/key.rs
```diff
@@ -85,7 +85,7 @@ impl RedisKey {
         let val = if self.is_null() {
             None
         } else {
-            hash_get_key(self.ctx, self.key_inner, field)
+            hash_mget_key(self.ctx, self.key_inner, &[field])?.pop().unwrap()
```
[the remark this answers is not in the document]
slavak: If the field doesn't exist `hash_mget_key` returns `None` in the corresponding (in this case, only) entry, so this will return `None` as well.

(I'm assuming you asked because of the `.unwrap()`? That's there because `pop()` returns an `Option` in case the vector doesn't have an element to pop. Since in this case we know the vector is exactly 1 element in size it should always be `Some`.)

src/raw.rs
slavak: If you ask me I say the more the merrier! For the record I did 12 because that's how many fit neatly into 2 lines. (One line would have given us 6, which is a little low.) :smiley:

src/key.rs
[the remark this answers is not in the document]
gavrie: > If the field doesn't exist `hash_mget_key` returns `None` in the corresponding (in this case, only) entry, so this will return `None` as well.

Good.
 
> (I'm assuming you asked because of the `.unwrap()`?

Indeed.

> That's there because `pop()` returns an `Option` in case the vector doesn't have an element to pop. Since in this case we know the vector is exactly 1 element in size it should always be `Some`.)

Very well; so please document that assumption in the code by using `expect(...)` with an appropriate message instead of `unwrap`.

src/key.rs
```diff
@@ -246,17 +246,34 @@ fn read_key(key: *mut raw::RedisModuleKey) -> Result<String, Utf8Error> {
...
+pub fn hash_mget_key<T>(ctx: *mut raw::RedisModuleCtx,
+                        key: *mut raw::RedisModuleKey,
+                        fields: &[T])
+    -> Result<Vec<Option<RedisString>>, RedisError>
+    where T: Into<Vec<u8>> + Clone
+{
+    use crate::redisraw::bindings::RedisModuleString;
+    const BATCH_SIZE: usize = 12;
+
+    let mut values: Vec<Option<RedisString>> = Vec::with_capacity(fields.len());
```
gavrie, line 260: Why the explicit type annotation on `values`?

src/raw.rs
```diff
@@ -232,18 +232,59 @@ pub fn string_dma(key: *mut RedisModuleKey, len: *mut size_t, mode: KeyMode) ->
...
+                         values: &mut [*mut RedisModuleString])
+    -> Result<(), RedisError>
+    where T: Into<Vec<u8>> + Clone
+{
+    assert_eq!(fields.len(), values.len());
+
+    let mut fi = fields.iter();
+    let mut vi = values.iter_mut();
+
+    macro_rules! RM {
```
gavrie, line 245: `RM` is not a valid name for a macro; should be lowercase/snake_case instead.
Are you using the IntelliJ Rust plugin? Or VS Code? These should report things like this.

src/raw.rs
```diff
@@ -232,18 +232,59 @@ pub fn string_dma(key: *mut RedisModuleKey, len: *mut size_t, mode: KeyMode) ->
...
-            key,
-            REDISMODULE_HASH_CFIELDS as i32,
-            CString::new(field).unwrap().as_ptr(),
-            &res,
-            ptr::null::<c_char>(),
-        );
+pub fn hash_get_multi<T>(key: *mut RedisModuleKey, fields: &[T],
+                         values: &mut [*mut RedisModuleString])
+    -> Result<(), RedisError>
+    where T: Into<Vec<u8>> + Clone
```
gavrie, line 238: @slavak what does the generic parameter `T` provide? Why not simply use `fields: &[&str]` instead of `&[T]`?
If you see a need to provide e.g. `&[&String]`, you can use `AsRef<str>` as a bound.

Using `&[&str]` also allows getting ride of the `clone()` call in `CString::new` below (provided you use `let mut fi = fields.into_iter()`).

slavak: `fields: &[&str]` would prevent you from passing in a `&[String]` without allocating a new vector, which is unnecessarily limiting. `Into<Vec<u8>> + Clone` is just the minimal bound we can specify here. (Required because that's the bound for `CString::new`.)

The `clone` is necessary because iterating over a slice returns a borrowed reference for both `iter` and `into_iter`, while `CString::new` takes ownership of its argument. For slices of `&str` this should be very cheap because `&str` is `Copy`. For slices of `String` it's unfortunately a necessary cost unless we want to take ownership of the data, but I'm not sure how to do that in a way that's ergonomic.

src/key.rs
slavak: Good point.

src/raw.rs
slavak: Incidentally for `&[&str]` the clone shouldn't do any additional work because it's equivalent to the copy that already happens implicitly.

slavak:
@gavrie Pushed fixes.

src/raw.rs
slavak: :+1:

src/key.rs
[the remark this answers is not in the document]
slavak: :+1:

src/raw.rs
gavrie: Sounds good to me.

gavrie (commented):
Looks good!

The only thing I would add is a comment that explains why we're doing all these contortions to call `HashGet`: That it's varargs, no standard way of calling, yada yada.
Also refer to the relevant Redis issue: https://github.com/redis/redis/issues/7860

slavak:
@gavrie Pushed a new commit with an explanation comment. Take a look.

gavrie (approved):

PR #106, opened 2020-10-06
Add API for context-less logging

slavak:
- It is sometimes useful to log messages to the Redis logs from code that does not take a context
  object. The Redis modules API allows for this by specifying a NULL context, but the Rust wrapper
  previously prevented taking advantaged of this ability.
- Added a log_* method for each supported log level to the context object.
- Used `strum` to make translating the LogLevel enum to a string more efficient. This now uses
  static strings generated at compile-time, rather than allocating and formatting them dynamically
  at runtime on each log call.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,12 +3,6 @@
 use std::os::raw::c_char;
 use std::str::Utf8Error;
 
-#[macro_use]
-extern crate bitflags;
-#[macro_use]
-extern crate enum_primitive_derive;
-#[macro_use(AsRefStr)]
-extern crate strum_macros;
 extern crate num_traits;
 
 use libc::size_t;
@@ -17,16 +11,16 @@ pub mod alloc;
 pub mod error;
 pub mod native_types;
 pub mod raw;
+use strum_macros::AsRefStr;
 pub mod rediserror;
 mod redismodule;
 pub mod redisraw;
 pub mod redisvalue;
 
-#[macro_use]
-mod macros;
 mod context;
 mod key;
 pub mod logging;
+mod macros;
 
 #[cfg(feature = "experimental-api")]
 pub use crate::context::thread_safe::ThreadSafeContext;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,6 +2,7 @@
 
 use std::os::raw::c_char;
 use std::str::Utf8Error;
+use strum_macros::AsRefStr;
 
 extern crate num_traits;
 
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -7,6 +7,8 @@ use std::str::Utf8Error;
 extern crate bitflags;
 #[macro_use]
 extern crate enum_primitive_derive;
+#[macro_use(AsRefStr)]
+extern crate strum_macros;
```
gavrie, line 11: I think `extern crate` with `macro_use` is deprecated, and the `strum_macros` docs seem to agree:
`use strum_macros::{AsRefStr};`

It would be nice to fix the other `extern crate`s at the same time.

src/lib.rs
```diff
@@ -39,11 +42,15 @@ pub use crate::redismodule::*;
...
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, AsRefStr)]
 pub enum LogLevel {
+    #[strum(to_string = "debug")]
     Debug,
+    #[strum(to_string = "notice")]
     Notice,
+    #[strum(to_string = "verbose")]
     Verbose,
+    #[strum(to_string = "warning")]
```
gavrie, line 53: Looks like you could replace all those `#[strum(to_string = "debug")]` lines with a single `#[strum(serialize_all = "snake_case")]` on the `enum`.

gavrie (changes requested):

src/lib.rs
slavak: Done. Pushed as a new commit.

src/lib.rs
slavak: Good call.

slavak:
Rebased on master and pushed some fixes.

src/lib.rs
```diff
@@ -17,16 +11,16 @@ pub mod alloc;
 pub mod error;
 pub mod native_types;
 pub mod raw;
+use strum_macros::AsRefStr;
```
gavrie, line 14: Nice, but it'd be even nicer to put all the `use` statements together at the top and not mixed with `mod` statements.

gavrie (dismissed):

src/lib.rs
slavak: Whoops. I was counting on the formatter to sort it out.

slavak:
It looks like I also accidentally included some debug garbage, so removed that and fixed the `use` placement.

gavrie (approved):

PR #107, opened 2020-10-06
Add support for running module unload code

slavak:
* Can now run de-init code, similarly to how init code can be specified for a module.
* Added an example module to demonstrate the functionality.
* Also, fixed paths in macros, so that they can compile regardless of which identifiers
  the caller has brought into scope.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -12,19 +12,20 @@ macro_rules! redis_command {
 
         /////////////////////
         extern "C" fn do_command(
-            ctx: *mut raw::RedisModuleCtx,
-            argv: *mut *mut raw::RedisModuleString,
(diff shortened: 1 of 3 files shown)


src/macros.rs
```diff
@@ -12,19 +12,20 @@ macro_rules! redis_command {
 
         /////////////////////
         extern "C" fn do_command(
-            ctx: *mut raw::RedisModuleCtx,
-            argv: *mut *mut raw::RedisModuleString,
+            ctx: *mut $crate::raw::RedisModuleCtx,
+            argv: *mut *mut $crate::raw::RedisModuleString,
```
gavrie, line 16: Very good.

gavrie (approved):

PR #113, opened 2020-10-25
Provide an IntoIterator for the multi-get results

slavak:
This is both more idiomatic and more flexible than the previous `into_list`/`into_map`. It is now possible to both immediately collect the results into the user's choice of collection, and to perform further manipulations on the results via iterator methods without materializing them into an intermediate collection and incurring allocation costs.

Since the resulting iterator iterates of (key,value) pairs, this comes at the small cost of needing to explicitly `map` the results to collect them into a vector of field-values only.

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -14,7 +14,6 @@ use crate::redismodule::REDIS_OK;
 use crate::RedisError;
 use crate::RedisResult;
 use crate::RedisString;
-use std::iter::FromIterator;
 
 /// `RedisKey` is an abstraction over a Redis key that allows readonly
 /// operations.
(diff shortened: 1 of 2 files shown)


src/key.rs
```diff
@@ -268,54 +267,87 @@ where
     values: Vec<Option<RedisString>>,
 }
 
-impl<'a, T> HMGetResult<'a, T>
+pub struct HMGetIter<'a, T>
```
gavrie, line 270: Does `HMGetIter` need to be `pub`? Isn't it accessed only via the iterator?

src/key.rs
```diff
@@ -268,54 +267,87 @@ where
...
+where
+    T: Into<Vec<u8>>,
+{
+    ai: std::slice::Iter<'a, T>,
+    bi: std::vec::IntoIter<Option<RedisString>>,
```
gavrie, line 275: The names `ai` and `bi` aren't very clear. I suggest calling them something like `field_iter` and `value_iter`.

src/key.rs
```diff
@@ -268,54 +267,87 @@ where
...
+}
+
+impl<'a, T> Iterator for HMGetIter<'a, T>
 where
     T: Into<Vec<u8>> + Clone,
 {
-    /// Convert the results into any list-like collection. The field values in the collection
-    /// appear in the same order as that in which the field names were provided to the multi-get
-    /// call. Fields which do not exist are set to `None`.
+    type Item = (T, String);
```
gavrie, line 282: Using a `String` for the value is limiting -- what if the value is not a valid UTF-8 string? In any case, the iteration and the string conversion seem to be separate concerns. Why not let the caller use `map` with whatever conversion they prefer? Also, it might be possible to use `Into` and then call `into()` on the value to let the caller specify the conversion.

gavrie (changes requested):

src/key.rs
slavak: Yes, precisely _because_ it is accessed via the iterator:

```
error[E0446]: private type `key::HMGetIter<'a, T>` in public interface
   --> src/key.rs:308:5
    |
270 | struct HMGetIter<'a, T>
    | - `key::HMGetIter<'a, T>` declared as private
...
308 |     type IntoIter = HMGetIter<'a, T>;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't leak private type
```

gavrie: 👍

src/key.rs
slavak: Fixed in latest commit.

src/key.rs
slavak: Point taken. I've pushed a commit that makes this more flexible. You can now iterate by any type that satisfies `RedisString: Into<T>`. (And added a corresponding conversion for `String` for convenience.)

gavrie (approved):

PR #117, opened 2020-12-09
Fix all loglevels mistakenly set to debug

socs:
Fix for #106 where log_notice(), log_verbose() and log_warning() all set log level to LogLevel::Debug.


slavak (approved):
This is an embarrassing mistake indeed. Thank you for the PR!

Issue #91, opened 2020-06-28
Expose the RedisModule_OnLoad argv vector 


matobet:
Currently there appears to be no way how the user-provided `init` function can access the arguments vector passed to the `MODULE LOAD mymod arg1 arg2 ...` that is present in the C-api `int RedisModule_OnLoad(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)`. 

Exposing this would also in Rust code would be useful to pass some static configuration to the module at load time.

PR #122, opened 2021-02-21
Make init/deinit callbacks use high-level API

matobet:
Both the `commands` and `event_handlers` macro callbacks are written in terms of the high-level API. I.e. user gets passed a parameter a "high-level" `Context` (instead of `raw::RedisModuleCtx` pointer), args converted to strings... return values (when necessary) properly handled .etc. `init` and `deinit` appear to be exceptions to this because they require user to write the callbacks in terms of `raw::` types and manually casting statuses to `c_int`... This PR attempts to achieve full consistency by making the init/deinit callback use "high-level" types (though at a cost of a breaking change) so modules will no longer have to mix-and-match "high-level" and "raw" terminology for various callbacks.

* Now a constructed Context reference is passed to init/deinit callbacks
  (instead of former raw::RedisModuleCtx) and they are expected to
  return Status (Ok|Err) instead of c_int.
* Exposed raw::State as $crate::State to make the init/deint not depend
  on any raw:: types.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ Cargo.lock
 # IntelliJ
 .idea
 src/redisraw/bindings.rs
+
+# VS Code
+.vscode
--- a/examples/load_unload.rs
+++ b/examples/load_unload.rs
@@ -1,13 +1,11 @@
 #[macro_use]
 extern crate redis_module;
 
-use redis_module::{raw, Context, LogLevel};
-use std::os::raw::c_int;
+use redis_module::{Context, LogLevel, Status};
 
 static mut GLOBAL_STATE: Option<String> = None;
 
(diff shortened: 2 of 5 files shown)


matobet:
Now that we are changing the signature for `init` we might as well use the opportunity and tackle #91 ... If you think this is a good idea I can also look into contributing this too.

gavrie:
@matobet Thanks for the PR! I'll look at the code and comment.
I don't see any problem per se with making a breaking change; doing #91 sounds good as well.

gavrie (approved):
Good work, thanks!

matobet:
Thanks, always happy to contribute! And thank you for the quick turnaround :+1:

PR #127, opened 2021-02-25
Allow the use of NextArg on Iter<&str> as well as Iter<String>

gavrie:
This makes it easier to write tests that naturally provide a `Vec<&str>` instead of a `Vec<String>` which requires manually calling `to_string()` on all items.

The disadvantage of this change is that unlike the previous approach, if we already have an owned `String`, we clone it anyway. 

There doesn't seem to be a way to get both advantages at the same time.


gavrie:
@gkorland @slavak I'd like to hear your thoughts on this.

slavak:
Looks reasonable to me.

There's a chance the compiler can elide the extra clone since the owned source String gets dropped immediately after, though I haven't tried to verify if this actually happens.

In any case I don't see a saner approach until specialization is available in Rust stable, at which point we can consider using different versions for `String` and `AsRef<str>`.

slavak (approved):

PR #129, opened 2021-03-01
examples/threads: Drop lock early

prashanthpai:
Demonstrate releasing the thread safe context lock as soon as we're done
accessing redis data structures. Earlier, the lock was being held until
the value was in scope, which included the sleep time.


gavrie:
Thanks @prashanthpai!

gavrie (approved):

PR #130, opened 2021-03-02
Add support to delete hash fields

prashanthpai:


src/raw.rs
```diff
@@ -348,6 +349,24 @@ pub fn hash_set(key: *mut RedisModuleKey, field: &str, value: *mut RedisModuleSt
...
 }
 
+pub fn hash_del_single(key: *mut RedisModuleKey, field: &str) -> Status {
+    let field = CString::new(field).unwrap();
+
+    // support to pass multiple fields is desired but it's complicated
+    // see https://github.com/redis/redis/issues/7860
+
+    unsafe {
+        RedisModule_HashSet.unwrap()(
```
prashanthpai, line 359: It seems like the `RedisModule_HashSet` C API has been overloaded with multiple things then merely setting a hash value.

I'm very new to Rust. I don't know where to draw the line between this project providing a user-friendly Rust API to modules vs this project being a very thin and pass-through wrapper around the C modules API. Does it make sense to extend `hash_set` to take a flags or options instead?

slavak: The aim of this module is to provide an idiomatic Rust API for the Redis modules API. Using flag arguments to fundamentally change the functionality of a method isn't very common even in C, doubly so in Rust; so I think the approach you took here is the right one.

One thing I would change is rename all the instances of `hash_del_single` to simply `hash_del`, to match the existing style of `hash_set`/`hash_get`.

Oh, and thank you for the PR!

src/raw.rs
```diff
@@ -156,6 +156,7 @@ extern "C" {
 ///////////////////////////////////////////////////////////////
 
 pub const FMT: *const c_char = b"v\0".as_ptr() as *const c_char;
+pub const REDISMODULE_HASH_DELETE: *const RedisModuleString = 1 as *const RedisModuleString;
```
gavrie, line 159: I tried to figure out why this was needed -- I expected `bindgen` to generate this automatically.
It turns out it doesn't because it can't parse the typecast, see https://github.com/rust-lang/rust-bindgen/issues/316.
Please add a comment here that mentions this.

prashanthpai:
@gavrie @slavak Thank you for the quick review. I've updated the PR.

gavrie (approved):

PR #132, opened 2021-03-04
Implement list operations

prashanthpai:


examples/lists.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    match src_key.list_pop_head() {
+        None => {
+            return Ok(RedisValue::Null);
+        }
+        Some(value) => {
+            let ret_cpy = value.clone();
+            dst_key.list_push_tail(value);
+            return Ok(RedisValue::BulkString(ret_cpy.into()));
+        }
```
gavrie, line 36: This can be simplified a bit:
```suggestion
        None => Ok(RedisValue::Null),
        Some(value) => {
            let ret_cpy = value.clone();
            dst_key.list_push_tail(value);
            Ok(RedisValue::BulkString(ret_cpy.into()))
        }
```

examples/lists.rs
```diff
@@ -0,0 +1,49 @@
...
+// Pops and returns the first element (head) of the list stored at 'source'
+// and pushes the element to the last position (tail) of the list stored at
+// 'destination'.
+fn lpoprpush(ctx: &Context, args: Vec<String>) -> RedisResult {
+    if args.len() != 3 {
+        return Err(RedisError::WrongArity);
+    }
+
+    let src_key = ctx.open_key_writable(&args[1]);
+    let dst_key = ctx.open_key_writable(&args[2]);
```
gavrie, line 17: It would be nice to reuse the existing argument parsing and error handling code, which is more idiomatic in Rust (since it unifies the parsing and error handling instead of separating them):
```suggestion
    let mut args = args.into_iter().skip(1);
    
    let src = args.next_string()?;
    let dst = args.next_string()?;

    let src_key = ctx.open_key_writable(&src);
    let dst_key = ctx.open_key_writable(&dst);
```

gavrie (changes requested):
Nice work! Just a few style issues in the example.

gavrie (approved):
LGTM

PR #133, opened 2021-03-08
add support for notify_keyspace_event

gkorland:


src/raw.rs
```diff
@@ -474,3 +474,10 @@ pub fn subscribe_to_server_event(
 ) -> Status {
     unsafe { RedisModule_SubscribeToServerEvent.unwrap()(ctx, event, callback).into() }
 }
+
+#[cfg(feature = "experimental-api")]
+pub fn notify_keyspace_event(ctx: *mut RedisModuleCtx, event_type: i32, event: &str, keyname: *mut RedisModuleString) {
```
oshadmi, line 479: For `event_type`, we can also consider using `NotifyEvent` struct in raw.rs in redismodule-rs?

oshadmi (approved):

src/raw.rs
gavrie: @oshadmi Sounds like a good idea!

gkorland: I thought about it, but I think you should be able to send a union of values

gavrie: @gkorland That's not a problem! We use the [bitflags](https://docs.rs/bitflags/1.2.1/bitflags/) crate which support this out of the box (e.g. `NotifyEvent::HASH | NotifyEvent::ZSET`).

gkorland: fixed

gavrie (approved):

PR #149, opened 2021-05-19
Expose fn verify_type as public (to be used by RedisJSON isJSON API)

oshadmi:


gkorland:
@oshadmi I think you should create a wrapper method in RedisKey and not make this method pub.
```
pub fn verify_type(&self, redis_type: &RedisType) -> Result<(), RedisError>
```

oshadmi:
@gkorland Please consider that if a `RedisModuleKey` already exists, wrapping it with a `RedisKey` struct will require to open it, since it will be closed when `RedisKey` is dropped (so we need to add to the ref count, since it will be decremented on drop), or require to avoid opening it, and add some logic to avoid closing it when dropped.
In addition, since `verify_type` is not a struct method (no `self` arg),
it can be more easily exposed since (1) it has some added-value on its own, and (2) it doesn't require any additional data besides its `RedisModuleKey` pointer arg.

What about making it an associated functions (put it in an `impl` block of `RediKey`) instead of a global function,
But since it is used both by `RediKey` and by `RedisKeyWritable`, we'll need to do something like
Call it `Self::verify_type` from methods of `RedisKey` struct and
Call it `RedisKey::verify_type` from methods of `RedisKeyWritable` struct
(or simpler - always call it as `RedisKey::verify_type`)
(how can I `use` it from RedisJSON?)

What do you think?

gkorland (approved):

gavrie (approved):

PR #157, opened 2021-06-16
Move args from Vec<String> to Vec<RedisString> to reduce allocations

gkorland:


src/macros.rs
```diff
@@ -146,12 +144,10 @@ macro_rules! redis_module {
...
 
             let context = $crate::Context::new(ctx);
-            let args = $crate::decode_args(argv, argc);
-
-            if args.is_err() { return raw::Status::Err as c_int }
+            let args = $crate::decode_args(ctx, argv, argc);
 
             $(
-                if $init_func(&context, &args.unwrap()) == $crate::Status::Err {
+                if $init_func(&context, &args) == $crate::Status::Err {
```
oshadmi, line 150: Change of order of potential failure due to none utf8 in args?
Used to return error before `init_func`.
Now could return the error when doing `next_redis_string` in the `init_func` (but some other logic could precede it in the `init_func`)
Do we care about such minor/rare behavior changes - I assume not.

oshadmi (approved):
LGTM 👍🏼 (just minor comments)

gavrie (changes requested):

src/redismodule.rs
```diff
@@ -126,20 +116,42 @@ impl RedisString {
...
+    pub fn into_string_lossy(&self) -> String {
         let mut len: libc::size_t = 0;
         let bytes = unsafe { raw::RedisModule_StringPtrLen.unwrap()(self.inner, &mut len) };
         let bytes = unsafe { slice::from_raw_parts(bytes as *const u8, len) };
         String::from_utf8_lossy(bytes).into_owned()
     }
 
+    pub fn parse_unsigned_integer(&self) -> Result<u64, RedisError> {
+        let val = self.parse_integer()?;
+        u64::try_from(val).map_err(|_e| RedisError::String(format!("Couldn't parse as integer")))
```
gavrie, line 136: We could improve the error here to be something like "couldn't parse negative number as unsigned integer".

src/redismodule.rs
```diff
@@ -126,20 +116,42 @@ impl RedisString {
...
+    }
+
+    pub fn parse_integer(&self) -> Result<i64, RedisError> {
+        let mut val: i64 = 0;
+        match raw::string_to_longlong(self.inner, &mut val) {
+            raw::Status::Ok => Ok(val),
+            raw::Status::Err => Err(RedisError::String(format!("Couldn't parse as integer"))),
```
gavrie, line 143: No need for the `format!`:
```suggestion
            raw::Status::Err => Err(RedisError::Str("Couldn't parse as integer")),
```

src/redismodule.rs
```diff
@@ -155,6 +167,19 @@ impl Drop for RedisString {
...
+impl Display for RedisString {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.into_string_lossy())
+    }
+}
+
+impl Borrow<str> for RedisString {
+    fn borrow(&self) -> &str {
+        // TODO remove unwrap()
+        self.try_as_str().unwrap()
```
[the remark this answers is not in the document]
gavrie: @oshadmi Of course it can -- the TODO should be completed, probably by using `from_utf8_lossy` or by returning a fixed placeholder value (if we want to keep performance known).

tests/integration.rs
```diff
@@ -1,18 +1,41 @@
...
+use utils::{get_redis_connection, start_redis_server_with_module};
 
 #[test]
-#[ignore] // test will be run explicitly by a script
 fn test_hello() -> Result<()> {
-    prepare_dirs()?;
-
-    let _guards = vec![start_redis_server_with_module("hello")?];
-    let mut con = get_redis_connection()?;
+    let _guards = vec![start_redis_server_with_module("hello", 6479)?];
```
gavrie, line 9: Instead of hardcoding the port which is error-prone, we should use a function that allocates ports (e.g. a simple static counter that increments).

But why is this needed? Assuming tests don't run in parallel, we can just reuse the same port every time, can't we?

tests/integration.rs
```diff
@@ -1,18 +1,41 @@
...
+    let mut con = get_redis_connection(6479)?;
 
     let res: Vec<i32> = redis::cmd("hello.mul").arg(&[3, 4]).query(&mut con)?;
     assert_eq!(res, vec![3, 4, 12]);
 
+    let res: Result<Vec<i32>, RedisError> =
+        redis::cmd("hello.mul").arg(&["3", "xx"]).query(&mut con);
+    if let Ok(_) = res {
+        return Err(anyhow::Error::msg("Shold return an error"));
```
gavrie, line 18:
```suggestion
        return Err(anyhow::Error::msg("Should return an error"));
```

gavrie (changes requested):

tests/integration.rs
gavrie: When running the tests, add `-- --test-threads 1` to the `cargo test` command line to run the tests sequentially. That's why I didn't support multiple ports originally.
Do you think it's worth it to run the tests in parallel? Might make it harder to debug, and as long as they're fast it's not a problem. This may be worth discussing as a matter of principle -- what do other modules do with integration tests, run them sequentially or simultaneously?

src/redismodule.rs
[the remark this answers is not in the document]
gkorland:
```
        self.try_as_str().unwrap_or_default()
```
Should do?

tests/integration.rs
gkorland: I guess we should support parallel testing just to be on the safe side, but let's do it in separate PR

src/redismodule.rs
```diff
@@ -15,37 +19,40 @@ pub const REDIS_OK: RedisResult = Ok(RedisValue::SimpleStringStatic("OK"));
...
 }
 
-impl<S, T> NextArg for T
+impl<T> NextArg for T
 where
-    T: Iterator<Item = S>,
-    S: AsRef<str>,
+    T: Iterator<Item = RedisString>,
 {
+    fn next_redis_string(&mut self) -> Result<RedisString, RedisError> {
```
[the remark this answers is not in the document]
gkorland:
```
 In fact, now that I think of it, maybe the other method names such as next_i64 don't make sense as well and we should use a generic implementation that's polymorphic on the return type
```
@gavrie I'm not how do you want to do it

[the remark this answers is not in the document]
gavrie: >  In fact, now that I think of it, maybe the other method names such as next_i64 
> don't make sense as well and we should use a generic implementation that's polymorphic on the return type
>> @gavrie I'm not how do you want to do it

I'm not 100% sure myself; will need to experiment a bit. In any case there's no reason to hold this PR up.

src/redismodule.rs
```diff
@@ -155,6 +168,19 @@ impl Drop for RedisString {
...
+impl Display for RedisString {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.into_string_lossy())
+    }
+}
+
+impl Borrow<str> for RedisString {
+    fn borrow(&self) -> &str {
+        // RedisString might not be UTF-8 safe
+        self.try_as_str().unwrap_or_default()
```
gavrie, line 180: The default may be a bit weird since it'll be empty. It might be better to predefine a placeholder string such as `"<Invalid UTF-8 data>"` and return that (via `impl Default`?).

tests/integration.rs
```diff
@@ -1,18 +1,47 @@
...
+        return Err(anyhow::Error::msg("Should return an error"));
+    }
+
+    Ok(())
+}
+
+#[test]
+fn test_keys_pos() -> Result<()> {
+    let _guards = vec![start_redis_server_with_module("keys_pos", 6480)?];
+    let mut con = get_redis_connection(6480)?;
```
gavrie, line 33: I suggest adding context here as well, like in the `hello` test.

gavrie (changes requested):

src/redismodule.rs
gkorland:
```
        self.try_as_str().unwrap_or_else("<Invalid UTF-8 data>")
```

gavrie (approved):

PR #161, opened 2021-07-08
add Buffer(Vec<u8>) support for RedisValue

gkorland:


src/redisvalue.rs
```diff
@@ -6,6 +6,7 @@ pub enum RedisValue {
     SimpleString(String),
     BulkString(String),
     BulkRedisString(RedisString),
+    Buffer(Vec<u8>),
```
gavrie, line 9: For consistency, I think we should call this `StringBuffer` instead of `Buffer`. We're talking about yet another way to represent a Redis String.

gavrie (changes requested):

src/context/mod.rs
```diff
@@ -157,6 +157,15 @@ impl Context {
                 raw::RedisModule_ReplyWithString.unwrap()(self.ctx, s.inner).into()
             },
 
+            Ok(RedisValue::StringBuffer(s)) => unsafe {
+                raw::RedisModule_ReplyWithStringBuffer.unwrap()(
+                    self.ctx,
+                    s.as_ptr() as *const i8,
+                    s.len() as usize,
```
oshadmi, line 164: Can Vec len and the corresponding StringBuffer len be different (when bytes count does not match the character count in utf-8)
No need to call `from_utf8` or `into_string_lossy` or some conversion/validation functions?

oshadmi (approved):
👍🏼

PR #162, opened 2021-07-11
Add next_str to NextArg

gkorland:
1. Add `next_str` to NextArg
2. mark all NextArg as `[#inline]`

### DIFF
--- a/src/redismodule.rs
+++ b/src/redismodule.rs
@@ -21,6 +21,7 @@ pub const TYPE_METHOD_VERSION: u64 = raw::REDISMODULE_TYPE_METHOD_VERSION as u64
 pub trait NextArg {
     fn next_arg(&mut self) -> Result<RedisString, RedisError>;
     fn next_string(&mut self) -> Result<String, RedisError>;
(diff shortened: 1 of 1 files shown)


src/redismodule.rs
```diff
@@ -31,31 +32,43 @@ impl<T> NextArg for T
...
+    #[inline]
     fn next_string(&mut self) -> Result<String, RedisError> {
         self.next()
             .map_or(Err(RedisError::WrongArity), |v| Ok(v.into_string_lossy()))
     }
 
+    #[inline]
+    fn next_str<'a>(&mut self) -> Result<&'a str, RedisError> {
+        self.next()
+            .map_or(Err(RedisError::WrongArity), |v| v.try_as_str())
```
oshadmi, line 49: `next_string` is doing a lossy conversion (calling indirectly to `String::from_utf8_lossy`)
And `next_str` is not (calling indirectly to `str::from_utf8`)
Is this intended?

gkorland: yes, I think we should reconsider the from_utf8_lossy

oshadmi (approved):
👍🏼

PR #163, opened 2021-07-11
Clippy pedantic

gkorland:


src/key.rs
```diff
@@ -462,6 +480,10 @@ fn to_raw_mode(mode: KeyMode) -> raw::KeyMode {
     }
 }
 
+/// # Panics
+///
+/// Will panic if `RedisModule_KeyType` or `RedisModule_ModuleTypeGetType` are missing in redismodule.h
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
```
oshadmi, line 486: Adding here `#[allow(clippy::not_unsafe_ptr_arg_deref)]` in order to to avoid adding `unsafe`, which then would also require adding `unsafe` to all callers recursively?
And since this would probably affect many callers and would require many unsafe blocks?
So we prefer to add this configuration and silence this error (`deny` level) since we know the low level APIs should exist?

src/raw.rs
```diff
@@ -343,10 +360,11 @@ where
 
     match res {
         Status::Ok => Ok(()),
-        _ => Err(RedisError::Str("ERR key is not a hash value")),
+        Status::Err => Err(RedisError::Str("ERR key is not a hash value")),
```
oshadmi, line 363: match_wildcard_for_single_variants - OK (good to know)

oshadmi (approved):
👍🏼 LGTM

PR #168, opened 2021-07-18
add support for Command filter

gkorland:


src/commandfilter/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+        let arg = unsafe {
+            raw::RedisModule_CommandFilterArgGet.unwrap()(self.ctx, pos as c_int)
+                as *mut raw::RedisModuleString
+        };
+        RedisString::new(ptr::null_mut(), arg)
+    }
+
+    pub fn args_insert(&self, pos: usize, arg: RedisString) -> usize {
+        // retain arg to since RedisModule_CommandFilterArgInsert going to release it too
+        raw::string_retain_string(std::ptr::null_mut(), arg.inner);
```
oshadmi, line 29: Is it certain that `arg.inner` would be released by redis?
(I do not see it directly done in `RM_CommandFilterArgInsert`)

In comment about `RM_CommandFilterArgInsert`
I see:
```
The specified RedisModuleString argument may be used by Redis
 * after the filter context is destroyed, so it must not be auto-memory
 * allocated, freed or used elsewhere.
```

And also please see comment about `args_replace` when `pos` is out of bounds.

gkorland: yes they will be released like every other arguments

src/macros.rs
```diff
@@ -82,6 +82,31 @@ macro_rules! redis_event_handler {
     }};
 }
 
+#[cfg(feature = "experimental-api")]
+#[macro_export]
+macro_rules! redis_commnad_filter {
```
gavrie, line 87: Typo:
```suggestion
macro_rules! redis_command_filter {
```

gavrie (changes requested):

PR #169, opened 2021-07-18
initial commit of dictionary support

gkorland:


gavrie:
@gkorland what's the point of using this dictionary interface in a Rust module? I understand it might be convenient when writing in C, but in Rust you can just use `HashMap` with the Redis allocator.

gavrie (commented):
@gkorland what's the point of using this dictionary interface in a Rust module? I understand it might be convenient when writing in C, but in Rust you can just use HashMap with the Redis allocator.

PR #177, opened 2021-07-28
Fix #171: use platform independent c_char type

### LINKED ISSUE #171: Compilation error on aarch64
I was trying to compile a module on a Raspberry Pi(`aarch64`) and I encountered the following error:
```bash 
  --> /srv/mcaptchaguard/.cargo/registry/src/github.com-1ecc6299db9ec823/redis-module-0.21.0/src/context/mod.rs:175:21
    |
175 |                     s.as_ptr() as *const i8,
    |                     ^^^^^^^^^^^^^^^^^^^^^^^ expected `u8`, found `i8`
    |
    = note: expected raw pointer `*const u8`
               found raw pointer `*const i8`

```
I applied the suggested fix and the compilation succeeded. 

[Rust doc](https://doc.rust-lang.org/nightly/std/os/raw/type.c_char.html) says `std::os::raw::c_char` can either be `i8` or `u8` but I don't understand why `i8` would fail to compile.

[Here's the module source code](https://github.com/mCaptcha/cache)

---
The same module compiles fine on `x86_64`.

gavrie:


gavrie:
@gkorland please take care to always use `c_char`, since signed/unsigned chars are platform dependent.

gkorland (approved):

PR #179, opened 2021-08-03
avoid string copy when sending reply

MeirShpilraien:


gkorland (approved):

src/context/mod.rs
```diff
@@ -158,9 +158,10 @@ impl Context {
             },
 
             Ok(RedisValue::BulkString(s)) => unsafe {
-                raw::RedisModule_ReplyWithString.unwrap()(
+                raw::RedisModule_ReplyWithStringBuffer.unwrap()(
                     self.ctx,
-                    RedisString::create(self.ctx, s.as_ref()).inner,
+                    s.as_ptr() as *const i8,
```
oshadmi, line 163: Better use platform independent types?
`s.as_ptr() as *const c_char`

MeirShpilraien: I tried it like this first but it did not compile for some reason, I can dig into it...

PR #184, opened 2021-08-10
Add get_redis_version command

oshadmi:


src/context/mod.rs
```diff
@@ -263,4 +266,54 @@ impl Context {
...
+            Ok(re) => match re.captures(s) {
+                Some(captures) => match captures.get(1) {
+                    Some(m) => Some(m.as_str()),
+                    None => None,
+                },
+                None => None,
+            },
+            Err(_) => None,
+        }
+    }
```
gavrie, line 282: No sure why you made this a method of `Context`...
In general, the version parsing is spread out over 3 functions: This one, `get_redis_version_internal` and `Version::from`.
The regex parsing also has some duplication.
What I suggest is: Write 1 function that parses the `info` response directly into a `Version`. You can use 3 capture groups to directly extract the major/minor/patch without later needing to split on `.`. This should be fairly simple and clean.

gavrie (changes requested):

tests/integration.rs
```diff
@@ -48,3 +49,29 @@ fn test_keys_pos() -> Result<()> {
...
+    let _guards = vec![start_redis_server_with_module("test_helper", port)
+        .with_context(|| "failed to start redis server")?];
+    let mut con =
+        get_redis_connection(port).with_context(|| "failed to connect to redis server")?;
+
+    let res: Vec<i64> = redis::cmd("test_helper.version")
+        .query(&mut con)
+        .with_context(|| "failed to run test_helper.version")?;
+    assert!(
+        (res[0] == 6 && res[1] == 2 && res[2] >= 3) || (res[0] == 6 && res[1] == 0 && res[2] >= 15)
```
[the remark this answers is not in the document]
oshadmi: Fixed to test for major > 0

Makefile
```diff
@@ -65,6 +65,7 @@ RUST_SOEXT.macos=dylib
 
 build:
 	cargo build --features experimental-api --all --all-targets  $(CARGO_FLAGS)
+	cargo build --features test --example test_helper --all-targets  $(CARGO_FLAGS)
```
gkorland, line 68:
```suggestion
	cargo build --features experimental-api,test --all --all-targets  $(CARGO_FLAGS)
```

build.sh
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env sh
 cargo build --features experimental-api --all --all-targets
+cargo build --features test --example test_helper --all-targets
```
gkorland, line 3:
```suggestion
cargo build --features experimental-api,test --all --all-targets
```

gkorland (changes requested):

gkorland (approved):

PR #185, opened 2021-08-10
CircleCI modifications

rafie:


.circleci/config.yml
```diff
@@ -211,18 +220,20 @@ on-integ-and-version-tags: &on-integ-and-version-tags
...
-          <<: *on-any-branch
+          <<: *not-on-integ-branch
       - platforms-build:
           context: common
+          #@@<<: *on-integ-and-version-tags
           <<: *on-any-branch
           matrix:
             parameters:
-              platform: [focal, bionic, xenial, centos7]
+              platform: [focal, bionic, xenial, centos8, centos7, buster]
```
chayim, line 237: IMHO we should remove all these random OSes. There isn't a behavioural difference between them. I realize that in the past we **wanted** this, but I think rust has proven stable enough.

PR #195, opened 2021-11-24
Expose `RedisModule_GetServerInfo` as a rust API

danni-m:


examples/info.rs
```diff
@@ -0,0 +1,33 @@
...
+    let mut args = args.into_iter().skip(1);
+
+    let section: RedisString = args.next_arg()?;
+    let field: RedisString = args.next_arg()?;
+
+    let server_info = ctx.server_info(section.borrow());
+    match server_info.field(field.borrow()) {
+        None => Ok(RedisValue::Null),
+        Some(v) => Ok(RedisValue::BulkRedisString(v)),
+    }
```
gavrie, line 21: 1. There's no need for the `borrow`, just use `next_str` instead of `next_arg`.
2. When matching on builtin types such as `Option`, it's almost always more idiomatic to use combinators rather than a raw `match`.

```suggestion
    let section = args.next_str()?;
    let field = args.next_str()?;

    let server_info = ctx.server_info(section);

    Ok(server_info
        .field(field)
        .map_or(RedisValue::Null, RedisValue::BulkRedisString))
```

gavrie (changes requested):

gavrie (approved):

PR #208, opened 2021-12-28
fix #207 add support for RedisModule_GetCurrentCommandName

### LINKED ISSUE #207: Add support for GetCurrentCommandName
```c
REDISMODULE_API const char *(*RedisModule_GetCurrentCommandName)(RedisModuleCtx *ctx) REDISMODULE_ATTR;
```

gkorland:


src/context/mod.rs
```diff
@@ -282,6 +282,15 @@ impl Context {
         raw::notify_keyspace_event(self.ctx, event_type, event, keyname)
     }
 
+    #[cfg(feature = "experimental-api")]
+    pub fn current_command_name(&self) -> &str {
+        unsafe {
+            CStr::from_ptr(raw::RedisModule_GetCurrentCommandName.unwrap()(self.ctx))
```
oshadmi, line 288: Will crash on redis prior to 6.2 
(https://github.com/redis/redis/pull/8792)
I suggest to check if available, and if not return "N/A" etc., or even better to allow it to fail.

MeirShpilraien (approved):

PR #211, opened 2022-01-09
Add optional info function to redis_module macro

oshadmi:
Adding a new argument to `redis_module` macro, which allows specifying a callback function for adding specific module information to the `INFO` command,
e.g., `INFO MODULES`,
or `INFO <module_name>`, for example:
```
127.0.0.1:6379> INFO hello
# hello_hello
hello_field:value
127.0.0.1:6379> 

```

(this is in addition to the already used callback function for adding stack trace on panic/crash)

### DIFF
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -1,10 +1,10 @@
 use std::ffi::CString;
-use std::os::raw::{c_char, c_int, c_long};
+use std::os::raw::{c_char, c_int, c_long, c_longlong};
 use std::ptr;
 
 use crate::key::{RedisKey, RedisKeyWritable};
 use crate::raw::ModuleOptions;
-use crate::{add_info_field_str, raw, Status};
+use crate::{add_info_field_long_long, add_info_field_str, raw, Status};
 use crate::{add_info_section, LogLevel};
 use crate::{RedisError, RedisResult, RedisString, RedisValue};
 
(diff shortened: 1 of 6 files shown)


src/context/mod.rs
```diff
@@ -286,3 +286,36 @@ impl Context {
         unsafe { raw::RedisModule_SetModuleOptions.unwrap()(self.ctx, options.bits()) };
     }
 }
+
+pub struct InfoContext {
+    pub ctx: *mut raw::RedisModuleInfoCtx,
```
gkorland, line 291: @oshadmi why `pub`?

src/macros.rs
```diff
@@ -113,7 +114,10 @@ macro_rules! redis_module {
             ctx: *mut $crate::raw::RedisModuleInfoCtx,
             for_crash_report: i32,
         ) {
-            $crate::base_info_func(ctx, for_crash_report == 1);
+            use $crate::InfoContext;
+            let mut __info_func__cb : Option<fn(&InfoContext, bool)> = None;
+            $( __info_func__cb = Some($info_func); )?
+            $crate::base_info_func(&$crate::InfoContext::new(ctx), for_crash_report == 1, __info_func__cb);
```
gkorland, line 120: @oshadmi no issues calling `base_info_func` when `__info_func__cb==None`?
Do we need to call `base_info_func` method if no `info_func` was passed?

gkorland (changes requested):

src/context/mod.rs
oshadmi: It is used outside, similar to `Context`, e.g., used by other modules, such as RedisJSON, etc.

src/macros.rs
oshadmi: We need `base_info_func` to add panic/crash info - always.
The user info is optional, and when it is `None`, no info is added, except for panic/crash info.

gkorland (approved):

PR #227, opened 2022-04-13
update cargo edition to 2021

gkorland:


matobet:
@oshadmi @gkorland are we ok to merge this?

PR #240, opened 2022-08-30
add ERR prefix to error

gkorland:


src/key.rs
```diff
@@ -502,7 +502,7 @@ pub fn verify_type(key_inner: *mut raw::RedisModuleKey, redis_type: &RedisType)
         let raw_type = unsafe { raw::RedisModule_ModuleTypeGetType.unwrap()(key_inner) };
 
         if raw_type != *redis_type.raw_type.borrow() {
-            return Err(RedisError::Str("Existing key has wrong Redis type"));
+            return Err(RedisError::Str("ERR Existing key has wrong Redis type"));
```
oshadmi, line 505: 1. I see several other similar errors without the `ERR` prefix
(searching for `Err(RedisError::Str(`)

2. Maybe we should have this `ERR` prefix added by `RedisError` (if not already present) and not hard-coded?
Similar to `impl<T: std::error::Error> From<T> for RedisError`

src/rediserror.rs
```diff
@@ -45,6 +45,6 @@ impl fmt::Display for RedisError {
             RedisError::String(s) => s.as_str(),
         };
 
-        write!(f, "{}", d)
+        write!(f, "ERR {d}")
```
oshadmi, line 48: Breaks test error::tests::test_to_redis_error

MeirShpilraien:
What if the user want different error code other then `ERR`?

MeirShpilraien:
What if we want different error code then ERR?

PR #241, opened 2022-08-30
Add support for unlink

gkorland:


src/key.rs
```diff
@@ -269,6 +269,14 @@ impl RedisKeyWritable {
         REDIS_OK
     }
 
+    /// # Panics
+    ///
+    /// Will panic if `RedisModule_UnlinkKey` is missing in redismodule.h
+    pub fn unlink(&self) -> RedisResult {
+        unsafe { raw::RedisModule_UnlinkKey.unwrap()(self.key_inner) };
```
oshadmi, line 276: [RedisModule_UnlinkKey](https://redis.io/docs/reference/modules/modules-api-ref/)
> On success REDISMODULE_OK is returned. If the key is not open for writing REDISMODULE_ERR is returned.

Consider returning the actual result

gkorland: Yes, thought about it, but for consistency I just copied the same API of delete

oshadmi: If we would like to fix both, when can we do such a breaking change in this module?
Since it may affect other consumers and not only RedisJSON, a config/feature-flag may help.

oshadmi (approved):
👍🏼

Issue #246, opened 2022-10-24
How to use binary/bytes/stringbuffer type


thomas9911:
I'm making a module that wants to read/write non-utf8 bytes to Redis. However the [RedisKeyWritable](https://docs.rs/redis-module/1.0.1/redis_module/key/struct.RedisKeyWritable.html#) read / write function only accept &str and also the higher level call function only accepts strings.

Is there a way to do this currently (without using the raw C bindings) ?

Currently i worked around this by creating my own type and using the set_value / get_value functions.

gkorland:
@thomas9911 you're right it seems like a left over from the original API.
The original API didn't work with byte safe RedisString but with String/str.
We should update the API to support RedisString (PR is welcome)

tBuLi12:
Since I also bumped into this missing functionality the other day I sketched a quick draft that just adds byte-slice-accepting versions of the original methods. 

Although that might work, perhaps a more generic interface should be implemented (tho even then something like 
`T: Into<Vec<u8>>` cannot replace `&str` in the existing api, since `&str` accepts `&&str` and `T: Into<Vec<u8>>` does not)

thomas9911:
I was also working on it 😅

tBuLi12:
Oh damn, sorry - well, mine was just a sketch and I see your version accepts RedisStrings rather than byte slices (as it should) so nvm that 😅 

Idk about breaking the public API, but keeping the default read & write with plain old strings as arguments does not seem like good long-term choice either...

tBuLi12:
Well we could make the api accept `T: Into<RedisString>` (at least in places where the underlying string is cloned anyway), and also
add an `impl Into<RedisString> for &str`, but I'm not sure if that keeps it completely backwards compatible

PR #248, opened 2022-10-25
feat: convert call and write functions to use RedisString

thomas9911:
based on https://github.com/RedisLabsModules/redismodule-rs/issues/246

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -1,7 +1,7 @@
 use std::convert::TryFrom;
+use std::ffi::{CString, NulError};
 use std::os::raw::c_void;
 use std::ptr;
-use std::str::Utf8Error;
 use std::time::Duration;
 
 use libc::size_t;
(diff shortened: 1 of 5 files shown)


src/key.rs
```diff
@@ -453,12 +456,13 @@ impl Drop for RedisKeyWritable {
     }
 }
 
-fn read_key(key: *mut raw::RedisModuleKey) -> Result<String, Utf8Error> {
+fn read_key(key: *mut raw::RedisModuleKey) -> Result<(CString, size_t), NulError> {
```
MeirShpilraien, line 459: Why not return a slice here?

gkorland:
@thomas9911 thanks for the PR.
I'm going to break this PR to two PRs one for `call` and the read will be handled differently

gkorland:
fixed in #264 #263

PR #253, opened 2022-12-22
cleaning clippy warnings

gkorland:


examples/data_type.rs
```diff
@@ -33,19 +33,25 @@ static MY_REDIS_TYPE: RedisType = RedisType::new(
...
+        copy2: None,
+        free_effort2: None,
+        mem_usage2: None,
+        unlink2: None,
     },
 );
 
 unsafe extern "C" fn free(value: *mut c_void) {
-    Box::from_raw(value.cast::<MyType>());
+    drop(Box::from_raw(value.cast::<MyType>()));
```
MeirShpilraien, line 46: Why do you need it here?

src/include/redismodule.h
```diff
@@ -4,6 +4,29 @@
 #include <sys/types.h>
 #include <stdint.h>
 #include <stdio.h>
+#include <stdlib.h>
```
MeirShpilraien, line 7: Did you took this file from unstable or from Redis 7.0 branch?

MeirShpilraien (commented):
Small questions..

examples/data_type.rs
gkorland:
```
$ cargo build --all --all-targets
   Compiling redis-module v99.99.99 (/home/guy/workspace/redismodule-rs)
warning: unused return value of `Box::<T>::from_raw` that must be used
  --> examples/data_type.rs:45:5
   |
45 |     Box::from_raw(value.cast::<MyType>());
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: call `drop(from_raw(ptr))` if you intend to drop the `Box`
   = note: `#[warn(unused_must_use)]` on by default
```

src/include/redismodule.h
gkorland: fixed

src/lib.rs
```diff
@@ -52,8 +50,8 @@ pub enum LogLevel {
     Warning,
 }
 
-fn from_byte_string(byte_str: *const c_char, length: size_t) -> Result<String, Utf8Error> {
-    let mut vec_str: Vec<u8> = Vec::with_capacity(length as usize);
+fn from_byte_string(byte_str: *const c_char, length: usize) -> Result<String, Utf8Error> {
```
oshadmi, line 53: So we can assume `size_t` is `usize` (except for niche platforms, which will break as explained [here](https://github.com/rust-lang/rust-bindgen/issues/1903) and [here](https://github.com/dtolnay/cxx/pull/576/files#diff-5e36084aec44d61b84d68bf00fcf75d7164d22496225c357d8f3a56b5f4768b5R261))

src/raw.rs
```diff
@@ -547,7 +547,7 @@ pub fn save_unsigned(rdb: *mut RedisModuleIO, val: u64) {
 }
 
 #[allow(clippy::not_unsafe_ptr_arg_deref)]
-pub fn string_compare(a: *const RedisModuleString, b: *const RedisModuleString) -> Ordering {
+pub fn string_compare(a: *mut RedisModuleString, b: *mut RedisModuleString) -> Ordering {
```
oshadmi, line 550: Isn't it strange a comparison function requires `mut` args? It seems to be dictated by the header file,
Also as in redis 7.0
https://github.com/redis/redis/blob/b73de0d9c244cae350b65a08dd7f75c2ae1fb54d/src/redismodule.h#L1048 (args are not `const`)

But it isn't so on redis unstable
https://github.com/redis/redis/blob/af0a4fe20771603f0eab75a1f60748d124cf33c3/src/redismodule.h#L1112
(args are `const`)

gkorland: right but we can't use the unstable yet

oshadmi (approved):
👍🏼

PR #262, opened 2023-01-18
Convert call to use RedisString

gkorland:
Based on #248 (only include call switch to RedisString)

### DIFF
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -95,14 +95,8 @@ impl Context {
         }
     }
 
-    pub fn call(&self, command: &str, args: &[&str]) -> RedisResult {
-        let terminated_args: Vec<RedisString> = args
-            .iter()
-            .map(|s| RedisString::create(self.ctx, s))
-            .collect();
-
-        let inner_args: Vec<*mut raw::RedisModuleString> =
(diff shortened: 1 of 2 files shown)


src/context/mod.rs
```diff
@@ -95,14 +95,8 @@ impl Context {
...
-    pub fn call(&self, command: &str, args: &[&str]) -> RedisResult {
-        let terminated_args: Vec<RedisString> = args
-            .iter()
-            .map(|s| RedisString::create(self.ctx, s))
-            .collect();
-
-        let inner_args: Vec<*mut raw::RedisModuleString> =
-            terminated_args.iter().map(|s| s.inner).collect();
+    pub fn call(&self, command: &str, args: &[&RedisString]) -> RedisResult {
+        let inner_args: Vec<*mut raw::RedisModuleString> = args.iter().map(|s| s.inner).collect();
```
MeirShpilraien, line 99: If RedisString was just holding the raw pointer then maybe we could avoid this.

MeirShpilraien (commented):
Another idea, lets make call a template function such that the template value implements into RedisString?

gkorland:
Replaced by #264

PR #263, opened 2023-01-18
Refactor string DMA

gkorland:
1. Encapsulate the string when opened for Write  
2. Encapsulate string resize on write or append 

ref #248

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -505,6 +505,10 @@ impl<'a> StringDMA<'a> {
     pub const fn len(&self) -> usize {
         self.buffer.len()
     }
+
+    pub const fn is_empty(&self) -> bool {
+        self.buffer.is_empty()
+    }
 }
 
 impl From<raw::Status> for Result<(), RedisError> {
--- a/src/key.rs
+++ b/src/key.rs
@@ -1,4 +1,6 @@
 use std::convert::TryFrom;
+use std::ops::Index;
+use std::ops::IndexMut;
 use std::os::raw::c_void;
 use std::ptr;
 use std::time::Duration;
@@ -73,7 +75,7 @@ impl RedisKey {
         self.key_inner == null_key
     }
 
-    pub fn read<'a>(&self) -> Result<Option<&'a [u8]>, RedisError> {
+    pub fn read(&self) -> Result<Option<&[u8]>, RedisError> {
         if self.is_null() {
             Ok(None)
         } else {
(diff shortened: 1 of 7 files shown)


src/key.rs
```diff
@@ -75,13 +73,20 @@ impl RedisKey {
         self.key_inner == null_key
     }
 
-    pub fn read(&self) -> Result<Option<String>, RedisError> {
-        let val = if self.is_null() {
-            None
+    pub fn read<'a>(&self) -> Result<Option<&'a [u8]>, RedisError> {
```
MeirShpilraien, line 76:
```suggestion
    pub fn read(&self) -> Result<Option<&'_ [u8]>, RedisError> {
```
Not sure whats the `'a` is

src/key.rs
```diff
@@ -437,6 +437,56 @@ where
...
+        let dma = raw::string_dma(key.key_inner, &mut length, raw::KeyMode::WRITE);
+        if dma.is_null() {
+            Err(RedisError::Str("Could not read key"))
+        } else {
+            let buffer = unsafe { std::slice::from_raw_parts_mut(dma.cast::<u8>(), length) };
+            Ok(StringDMA { key, buffer })
+        }
+    }
+
+    pub fn write(&mut self, data: &[u8]) -> Result<&mut Self, RedisError> {
```
MeirShpilraien, line 457: Why not implement `std::ops::index`and `std::ops::index_mut`?
https://doc.rust-lang.org/std/ops/trait.Index.html
https://doc.rust-lang.org/std/ops/trait.IndexMut.html

src/key.rs
```diff
@@ -437,6 +437,56 @@ where
...
+            let dma = raw::string_dma(self.key.key_inner, &mut length, raw::KeyMode::WRITE);
+            self.buffer = unsafe { std::slice::from_raw_parts_mut(dma.cast::<u8>(), length) };
+        } else {
+            return Err(RedisError::Str("Failed to truncate string"));
+        }
+        self.buffer[current_len..new_len].copy_from_slice(data);
+        Ok(self)
+    }
+
+    pub fn as_bytes(&self) -> &[u8] {
```
MeirShpilraien, line 485: We can implement Deref
https://doc.rust-lang.org/std/ops/trait.Deref.html

MeirShpilraien (commented):
Small suggestions

src/key.rs
gkorland: It's an idea, but it can't be as efficient as a bulk write, I'll add it anyway.

src/key.rs
gkorland: do you need Deref? there's nothing to release here. StringDMA only holds borrows and its life time depends on the KeyHolder lifetime

MeirShpilraien: If you implement deref, you do not need to implement is empt ly (for example)

gkorland: is_empty is just for users simplicity

src/key.rs
gkorland: Any added Index and IndexMut too

MeirShpilraien (approved):

PR #266, opened 2023-01-23
Added API to read and trim a stream.

MeirShpilraien:
Taken from: https://github.com/RedisLabsModules/redismodule-rs/pull/229
Added API for stream integration.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -21,12 +21,20 @@ impl<'key> StreamIterator<'key> {
         mut from: Option<raw::RedisModuleStreamID>,
         mut to: Option<raw::RedisModuleStreamID>,
         exclusive: bool,
+        reverse: bool,
     ) -> Result<StreamIterator, RedisError> {
-        let flags = if exclusive {
+        let mut flags = if exclusive {
             raw::REDISMODULE_STREAM_ITERATOR_EXCLUSIVE as i32
         } else {
             0
         };
+
+        flags |= if reverse {
+            raw::REDISMODULE_STREAM_ITERATOR_REVERSE as i32
+        } else {
+            0
+        };
+
         let res = unsafe {
             raw::RedisModule_StreamIteratorStart.unwrap()(
                 key.key_inner,
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -1,8 +1,8 @@
-use crate::Status;
 use crate::key::RedisKey;
 use crate::raw;
 use crate::RedisError;
 use crate::RedisString;
+use crate::Status;
 use std::os::raw::c_long;
 use std::ptr;
 
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -34,10 +34,10 @@ impl<'key> StreamIterator<'key> {
                 to.as_mut().map_or(ptr::null_mut(), |v| v),
             )
         };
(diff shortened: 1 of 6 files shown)


src/stream.rs
```diff
@@ -0,0 +1,85 @@
...
+pub struct StreamIterator<'key> {
+    key: &'key RedisKey,
+}
+
+impl<'key> StreamIterator<'key> {
+    pub(crate) fn new(
+        key: &RedisKey,
+        mut from: Option<raw::RedisModuleStreamID>,
+        mut to: Option<raw::RedisModuleStreamID>,
+        exclusive: bool,
```
gkorland, line 22: @MeirShpilraien what about REDISMODULE_STREAM_ITERATOR_REVERSE?

src/stream.rs
```diff
@@ -0,0 +1,85 @@
...
+                flags,
+                from.as_mut().map_or(ptr::null_mut(), |v| v),
+                to.as_mut().map_or(ptr::null_mut(), |v| v),
+            )
+        };
+        if res != raw::REDISMODULE_OK as i32 {
+            Err(RedisError::Str("Failed creating stream iterator"))
+        } else {
+            Ok(StreamIterator { key })
+        }
```
gkorland, line 41:
```suggestion
if Status::Ok == res.into() {
            Ok(StreamIterator { key })
        } else {
            Err(RedisError::Str("Failed creating stream iterator"))
        }
```

src/stream.rs
```diff
@@ -0,0 +1,85 @@
...
+        let mut num_fields: c_long = 0;
+        let mut field_name: *mut raw::RedisModuleString = ptr::null_mut();
+        let mut field_val: *mut raw::RedisModuleString = ptr::null_mut();
+        if unsafe {
+            raw::RedisModule_StreamIteratorNextID.unwrap()(
+                self.key.key_inner,
+                &mut id,
+                &mut num_fields,
+            )
+        } != raw::REDISMODULE_OK as i32
```
gkorland, line 59:
```suggestion
        if Status::Ok
            != unsafe {
                raw::RedisModule_StreamIteratorNextID.unwrap()(
                    self.key.key_inner,
                    &mut id,
                    &mut num_fields,
                )
            }
            .into()
```

src/stream.rs
```diff
@@ -0,0 +1,85 @@
...
+            return None;
+        }
+        let mut fields = Vec::new();
+        while unsafe {
+            raw::RedisModule_StreamIteratorNextField.unwrap()(
+                self.key.key_inner,
+                &mut field_name,
+                &mut field_val,
+            )
+        } == raw::REDISMODULE_OK as i32
```
gkorland, line 70:
```suggestion
        while Status::Ok
            == unsafe {
                raw::RedisModule_StreamIteratorNextField.unwrap()(
                    self.key.key_inner,
                    &mut field_name,
                    &mut field_val,
                )
                .into()
            }
```

gkorland (changes requested):

src/stream.rs
MeirShpilraien: Good point, will add it.

MeirShpilraien: Done

gkorland (approved):

PR #278, opened 2023-02-07
fix build of code comments

gkorland:


oshadmi (approved):
We can add
```
cargo test --doc --package redis-module --features test
```
Ref https://github.com/rust-lang/cargo/issues/6669

PR #300, opened 2023-03-28
Allow the crate to operate when the Redis API isn't available.

iddm:
Such situations occur when, for example, the crate which uses the redismodule-rs crate as a dependency is run for testing, so without Redis available. In this case, there is no allocator available as well, and it will lead to panic. To enable both situations to work correctly under all circumstances, this change introduces a fallback mechanism back to the system allocator (default Rust allocator), which is always available.

Note that the changes are zero-cost: the code before these changes used unwrap(), which also had checks for the value being non-null and would panic at runtime if it was null; the current code does this check as well, but instead of panicking reverts back to the system allocator.

### DIFF
--- a/examples/test_helper.rs
+++ b/examples/test_helper.rs
@@ -12,7 +12,6 @@ fn test_helper_version(ctx: &Context, _args: Vec<RedisString>) -> RedisResult {
     Ok(response.into())
 }
 
-#[cfg(feature = "test")]
 fn test_helper_version_rm_call(ctx: &Context, _args: Vec<RedisString>) -> RedisResult {
     let ver = ctx.get_redis_version_rm_call()?;
     let response: Vec<i64> = vec![ver.major.into(), ver.minor.into(), ver.patch.into()];
@@ -43,7 +42,6 @@ fn add_info(ctx: &InfoContext, _for_crash_report: bool) {
 
 //////////////////////////////////////////////////////
 
-#[cfg(feature = "test")]
 redis_module! {
     name: "test_helper",
     version: 1,
--- a/src/alloc.rs
+++ b/src/alloc.rs
@@ -2,28 +2,30 @@ use std::alloc::{GlobalAlloc, Layout};
 
 use crate::raw;
 
+/// Panics with a message without using an allocator.
+/// Useful when using the allocator should be avoided or it is
+/// inaccessible. The default [std::panic] performs allocations and so
+/// will cause a double panic without a meaningful message if the
+/// allocator can't be used. This function makes sure we can panic with
+/// a reasonable message even without the allocator working.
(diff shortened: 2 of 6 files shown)


src/alloc.rs
```diff
@@ -1,29 +1,58 @@
...
 use crate::raw;
 
-pub struct RedisAlloc;
+/// Defines the Redis allocator. This allocator delegates the allocation
+/// and deallocation tasks to the Redis server when available, otherwise
+/// it fallbacks to the default Rust [std::alloc::System] allocator
+/// which is always available compared to the Redis allocator.
+#[derive(Copy, Clone)]
+pub struct RedisAlloc {
+    system: std::alloc::System,
```
gkorland, line 11: why do you need to keep it as a field?

src/alloc.rs
```diff
@@ -1,29 +1,60 @@
...
-use std::os::raw::c_void;
 
 use crate::raw;
 
+/// Panics with a message without using an allocator.
+/// Useful when using the allocator should be avoided or it is
+/// inaccessible. The default [std::panic] performs allocations and so
+/// will cause a double panic without a meaningful message if the
+/// allocator can't be used. This function makes sure we can panic with
+/// a reasonable message even without the allocator working.
```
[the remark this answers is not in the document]
oshadmi: @vityafx If Redis is not available, and we use the feature `fallback_to_system_allocator` then we will not crash due to a missing memory allocator, however what about other apis that are required from Redis - they would also be missing.
So this fallback could be useful for tests which do not require Redis apis? (or maybe have some Redis mock in place?)

examples/test_helper.rs
```diff
@@ -12,7 +12,6 @@ fn test_helper_version(ctx: &Context, _args: Vec<RedisString>) -> RedisResult {
     Ok(response.into())
 }
 
-#[cfg(feature = "test")]
```
oshadmi, line 15: This is for unit testing (to test internal implementation) - seems the feature `test` is still needed for this one?

oshadmi: Oh right, this is an example and anyway not going to be part of the crate binary.
Thanks.

oshadmi (approved):

BruAPAHE:
@vityafx @gkorland I get `Critical error: the Redis Allocator isn't available` panic when i run tests `cargo test`. Anybody resolve this problem?

MeirShpilraien:
Hey @BruAPAHE , look how we solve it on RedisGears: https://github.com/RedisGears/RedisGears/blob/c6de5d50eb92ecbdfc976fb5b15bd6beb9d44c12/redisgears_core/src/lib.rs#L1256

BruAPAHE:
@MeirShpilraien Thank you so much!!!

PR #307, opened 2023-04-06
Let the redis module implementation register the allocator.

iddm:
This fixes lots of issues which were present due to a global allocator being set forcefully by this crate. This is undesirable for the modules dependent on this crate, as this becomes trouble when running unit tests, for example, and requires falling back to the system allocator. As long as the Redis allocator is only required when running the library as a redis module rather than always, setting it globally should be avoided.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ endif
 test: cargo_test
 
 cargo_test:
-	cargo test --all-targets --features test,experimental-api $(CARGO_FLAGS)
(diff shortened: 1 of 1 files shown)


MeirShpilraien (commented):
@vityafx look good, but I believe we should stop maintaining this branch. Let's make this PR directly to the master  branch? WDYT?

MeirShpilraien:
@vityafx I guess we can close this one as this replaces it: https://github.com/RedisLabsModules/redismodule-rs/pull/309

PR #309, opened 2023-04-14
Let the redis module implementation register the allocator.

iddm:
This is a port of https://github.com/RedisLabsModules/redismodule-rs/pull/307 into master.

This fixes lots of issues which were present due to a global allocator being set forcefully by this crate. This is undesirable for the modules dependent on this crate, as this becomes trouble when running unit tests, for example, and requires falling back to the system allocator. As long as the Redis allocator is only required when running the library as a redis module rather than always, setting it globally should be avoided.

### DIFF
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -2,8 +2,9 @@ use crate::context::thread_safe::RedisLockIndicator;
 use crate::{raw, CallOptionResp, CallOptionsBuilder, CallResult, RedisGILGuard, RedisValue};
 use crate::{Context, RedisError, RedisString};
 use bitflags::bitflags;
-use std::ffi::{c_char, c_int, c_longlong, c_void, CStr, CString};
+use std::ffi::{CStr, CString};
 use std::marker::PhantomData;
+use std::os::raw::{c_char, c_int, c_longlong, c_void};
 use std::sync::atomic::{AtomicBool, AtomicI64, Ordering};
 use std::sync::Mutex;
 
--- a/src/context/call_reply.rs
+++ b/src/context/call_reply.rs
@@ -1,6 +1,6 @@
 use core::slice;
+use std::os::raw::c_char;
 use std::{
-    ffi::c_char,
     fmt,
     fmt::{Debug, Display, Formatter},
     marker::PhantomData,


tests/integration.rs
```diff
@@ -195,63 +198,6 @@ fn test_scan() -> Result<()> {
...
-        .query(&mut con)
-        .with_context(|| "failed to add data to the stream")?;
-
-    assert_eq!(res, 0);
-
-    Ok(())
-}
-
-#[test]
-fn test_call() -> Result<()> {
```
MeirShpilraien, line 239: Was this test moved somewhere? Why was it deleted?

tests/integration.rs
```diff
@@ -195,63 +198,6 @@ fn test_scan() -> Result<()> {
     Ok(())
 }
 
-#[test]
-fn test_stream_reader() -> Result<()> {
```
MeirShpilraien, line 199: Why deleting this test?

tests/integration.rs
```diff
@@ -149,8 +151,9 @@ fn test_test_helper_err() -> Result<()> {
     Ok(())
 }
 
-#[test]
-fn test_string() -> Result<()> {
+#[cfg(feature = "experimental-api")]
+#[cfg_attr(feature = "experimental-api", test)]
+fn test_stream_reader() -> Result<()> {
```
MeirShpilraien, line 156: Why renaming this test? Also, why the `#[test]` is not needed?

tests/integration.rs
```diff
@@ -111,8 +112,9 @@ fn test_command_name() -> Result<()> {
     Ok(())
 }
 
-#[test]
-fn test_test_helper_info() -> Result<()> {
+#[cfg(feature = "experimental-api")]
+#[cfg_attr(feature = "experimental-api", test)]
+fn test_helper_info() -> Result<()> {
```
MeirShpilraien, line 117: Why `#[test]` is not needed?

tests/integration.rs
```diff
@@ -74,7 +75,7 @@ fn test_test_helper_version() -> Result<()> {
 }
 
 #[cfg(feature = "experimental-api")]
-#[test]
+#[cfg_attr(feature = "experimental-api", test)]
```
MeirShpilraien, line 78: Why #[test] is not needed?

tests/integration.rs
```diff
@@ -51,8 +51,9 @@ fn test_keys_pos() -> Result<()> {
     Ok(())
 }
 
-#[test]
-fn test_test_helper_version() -> Result<()> {
+#[cfg(feature = "experimental-api")]
+#[cfg_attr(feature = "experimental-api", test)]
+fn test_helper_version() -> Result<()> {
```
MeirShpilraien, line 56: Why #[test] is not needed?

MeirShpilraien: Oh you put it on the `cfg_attr` :+1: , So we we also need `#[cfg(feature = "experimental-api")]`?

tests/integration.rs
MeirShpilraien: Moved to `cfg_attr`

tests/integration.rs
MeirShpilraien: Moved to cfg_attr

tests/integration.rs
MeirShpilraien: Moved to cfg_attr

MeirShpilraien (commented):
Small comment about the tests changes, the other changes looks good :+1:

MeirShpilraien (approved):

tests/integration.rs
MeirShpilraien: Why this test was renamed?

MeirShpilraien (changes requested):
One last comment on test renaming

PR #312, opened 2023-04-18
Delete all features

iddm:


build.rs
```diff
@@ -41,31 +41,16 @@ fn main() {
...
-
     build
         .file("src/redismodule.c")
         .include("src/include/")
         .compile("redismodule");
 
-    let mut build = bindgen::Builder::default();
-
-    if experimental_api() {
-        build = build.clang_arg(format!("-D{EXPERIMENTAL_API}").as_str());
```
MeirShpilraien, line 65: I think we actually need this, otherwise the API which is marked as experimental on redismodule.h will not be available.

src/context/mod.rs
```diff
@@ -11,28 +11,19 @@ use crate::{add_info_field_long_long, add_info_field_str, raw, utils, Status};
...
 use std::ffi::CStr;
 
 use self::call_reply::CallResult;
 use self::thread_safe::RedisLockIndicator;
 
-#[cfg(feature = "experimental-api")]
 mod timer;
 
-#[cfg(feature = "experimental-api")]
-pub mod thread_safe;
```
MeirShpilraien, line 24: Why removing this?

MeirShpilraien: Ohh it move down :+1:

MeirShpilraien (approved):
Looks good :+1: (some small comments)
@gkorland please approve that you are OK with removing this experiment API feature. We think it does more damage then good and all the modules enables it anyway. Also there is no clear path from been experiment feature to feature stability (on Redis itself) so we do not see a need to maintain this.

build.rs
MeirShpilraien: Yes right

MeirShpilraien (approved):

MeirShpilraien (approved):

PR #318, opened 2023-05-01
Provide the standard logging facility implementation.

iddm:
As the crate is mostly used as a redis module, the logging will most likely be only useful when printed on the server. As many rustaceans are familiar and used to using the de-facto standard "log" crate, it makes sense to provide the logging implementation for this crate to enable the most-commonly used way of logging messages in Rust.

The use of the facility is optional. To register the logging backend, it is required to call the `redis_module::logging::standard_log_implementation::setup()` method, after that all the logs will be directed to the already existing implementations:
```rust
// In RedisGears 2:
    let globals = get_globals_mut();
    globals.pool = Some(Mutex::new(ThreadPool::new(
        (*EXECUTION_THREADS.lock(ctx)) as usize,
    )));

    if let Err(e) = redis_module::logging::standard_log_implementation::setup() {
        ctx.log_warning(&format!("Couldn't setup the standard logging: {e}"));
    }

    log::info!("Test warn message");

    Status::Ok
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,6 +41,12 @@ pub use crate::raw::*;
 pub use crate::redismodule::*;
 use backtrace::Backtrace;
 
+/// The detached Redis module context (the context of this module). It
+/// is only set to a proper value after the module is initialised via the
+/// provided [redis_module] macro.
+/// See [DetachedContext].
+pub static MODULE_CONTEXT: DetachedContext = DetachedContext::new();
+
 pub fn base_info_func(
     ctx: &InfoContext,
     for_crash_report: bool,
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -17,12 +17,27 @@ pub enum RedisLogLevel {
     Warning,
 }
 
-pub(crate) fn log_internal(ctx: *mut raw::RedisModuleCtx, level: RedisLogLevel, message: &str) {
+impl From<log::Level> for RedisLogLevel {
+    fn from(value: log::Level) -> Self {
+        match value {
+            log::Level::Error | log::Level::Warn => Self::Warning,
+            log::Level::Info => Self::Notice,
+            log::Level::Debug => Self::Debug,
+            log::Level::Trace => Self::Verbose,
+        }
+    }
+}
+
+pub(crate) fn log_internal<L: Into<RedisLogLevel>>(
+    ctx: *mut raw::RedisModuleCtx,
+    level: L,
+    message: &str,
+) {
     if cfg!(test) {
         return;
     }
 
-    let level = CString::new(level.as_ref()).unwrap();
+    let level = CString::new(level.into().as_ref()).unwrap();
     let fmt = CString::new(message).unwrap();
     unsafe {
         raw::RedisModule_Log.expect(NOT_INITIALISED_MESSAGE)(ctx, level.as_ptr(), fmt.as_ptr())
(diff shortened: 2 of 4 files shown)


src/logging.rs
```diff
@@ -1,53 +1,122 @@
...
+
+/// `LogLevel` is a level of logging to be specified with a Redis log directive.
+#[derive(Clone, Copy, Debug, AsRefStr)]
+#[strum(serialize_all = "snake_case")]
+pub enum RedisLogLevel {
+    Debug,
+    Notice,
+    Verbose,
+    Warning,
+    Error,
```
MeirShpilraien, line 16: Did you test it? AFAIR Redis does not have error log level.

MeirShpilraien: Yes, warning is our best option

src/logging.rs
```diff
@@ -1,53 +1,143 @@
...
+    ///
+    /// This function may be called on a module startup, within the
+    /// module initialisation function (specified in the
+    /// [crate::redis_module] as the `init` argument, which will be used
+    /// for the module initialisation and will be passed to the
+    /// [raw::Export_RedisModule_Init] function when loading the
+    /// module).
+    #[allow(dead_code)]
+    pub fn setup() -> Result<(), SetLoggerError> {
+        log::set_logger(&LOGGER).map(|()| log::set_max_level(log::LevelFilter::Trace))
```
MeirShpilraien, line 109: So basically, by logging using a NULL context we will not get the module name as part of the log message, but if we use detached context to log Redis will print the module name as part of the log message. My suggestion, why not passing the detached context here to the setup and save it on the static logger, then we can use it when logging. WDYT?

src/logging.rs
```diff
@@ -1,53 +1,143 @@
...
+        }
+
+        fn log(&self, record: &Record) {
+            if !self.enabled(record.metadata()) {
+                return;
+            }
+
+            let message = format!(
+                "'{}' {}:{}: {}",
+                record.module_path().unwrap_or_default(),
```
MeirShpilraien, line 124: If we will take the suggestion I made above then I guess we do not need to print the module here because Redis will do it for us.

MeirShpilraien (commented):
Looks good :+1:, small suggestion.

src/logging.rs
MeirShpilraien: > The logging here may come from any crate, not just ours. If we use a dependency which also uses the log crate, this handler will be invoked, and so the module_path() will be that crate which logs

Ohh ok, in that case lets keep it.

src/macros.rs
```diff
@@ -173,6 +173,11 @@ macro_rules! redis_module {
             argv: *mut *mut $crate::raw::RedisModuleString,
             argc: std::os::raw::c_int,
         ) -> std::os::raw::c_int {
+            unsafe {
+                // We trust that Redis won't give us a null pointer as
+                // the module context pointer.
+                $crate::MODULE_CONTEXT = std::ptr::NonNull::new(ctx);
```
MeirShpilraien, line 179: This ctx here is located on the stack and will be gone once the OnLoad function will finished. We should create a detached context here and use it instead. Or better, why not give the logger `setup` function the detached context that should be created by the user?

src/lib.rs
```diff
@@ -44,15 +41,11 @@ pub use crate::raw::*;
...
-    Debug,
-    Notice,
-    Verbose,
-    Warning,
-}
+/// The detached Redis module context (the context of this module). It
+/// is only set to a proper value after the module is initialised via the
+/// provided [redis_module] macro.
+/// See [DetachedContext].
+pub static mut MODULE_CONTEXT: DetachedContext = DetachedContext::new();
```
MeirShpilraien, line 48: Do we need it to be mutable?

MeirShpilraien (approved):
:+1: 
One small question, not a blocker.

MeirShpilraien (approved):

PR #324, opened 2023-05-04
Fix publish CI.

MeirShpilraien:
Added `extra-traits` feature to `syn` dependency so we will get the Debug trait implementation.


MeirShpilraien:
Should fix this failure: https://github.com/RedisLabsModules/redismodule-rs/actions/runs/4879208532/jobs/8705557599

MeirShpilraien:
> I wonder how has it ever been compiled

Yes I also wonder the same thing. If you compile the crate itself it does not compile. But I guess we never really tried it and we just use it on the `redismodule-rs` crate which does seem to work.

PR #327, opened 2023-05-09
Use `BTreeMap` and `BTreeSet` in addition to `HashMap` and `HashSet`.

MeirShpilraien:
The main problem with `HashMap` and `HashSet` is that those are unordered. So we get results in different order each time.

Using `BTreeMap` and `BTreeSet` we will get the results in the same order each time.


MeirShpilraien:
We need to decide if we can do this change of a patch version.

Issue #331, opened 2023-05-15
Add a valgrind/memory sanitiser CI


MeirShpilraien:
We should have a CI that verify not leaks or un-allow memory access. We have a lot of unsafe code in this crate and we should make sure we are not doing mistakes.

PR #352, opened 2023-06-21
Fix lifetime warning in Rust 1.70

slavak:
Starting with rustc 1.70.0, the code in raw.rs:hash_get_multi produced the following warnings:

    this `CString` is deallocated at the end of the statement, bind it to a
    variable to extend its lifetime

While it would work in practice due to the underlying implementation just doing pointer juggling, this warning was correct: The CString created within the macro `f` would be destroyed before the reference to the pointer taken from it with `as_ptr` was used.

Resolved by allocating a new Vec that contains the CStrings. This ensures they remain alive until the function returns.

### DIFF
--- a/src/raw.rs
+++ b/src/raw.rs
@@ -473,7 +473,7 @@ where
 {
     assert_eq!(fields.len(), values.len());
 
-    let mut fi = fields.iter();
+    let mut fi = fields.iter().map(|s| CString::new(s.clone()).unwrap());
     let mut vi = values.iter_mut();
 
     macro_rules! rm {
@@ -491,9 +491,7 @@ where
     }
     macro_rules! f {
         () => {
-            CString::new((*fi.next().unwrap()).clone())
-                .unwrap()
-                .as_ptr()
+            fi.next().unwrap().as_ptr()
         };
     }
     macro_rules! v {
--- a/src/raw.rs
+++ b/src/raw.rs
@@ -473,7 +473,12 @@ where
 {
     assert_eq!(fields.len(), values.len());
 
-    let mut fi = fields.iter().map(|s| CString::new(s.clone()).unwrap());
+    let fields = fields
+        .iter()
+        .map(|e| CString::new(e.clone()))
+        .collect::<Result<Vec<CString>, _>>()?;
+
+    let mut fi = fields.iter();
     let mut vi = values.iter_mut();
 
     macro_rules! rm {


src/raw.rs
```diff
@@ -473,7 +473,7 @@ where
 {
     assert_eq!(fields.len(), values.len());
 
-    let mut fi = fields.iter();
+    let mut fi = fields.iter().map(|s| CString::new(s.clone()).unwrap());
```
MeirShpilraien, line 476: Are we sure there are no internal 0 in the data? Otherwise the unwrap here is wrong. We can either raise the error or (if we are sure) then maybe use `expect`?

src/raw.rs
```diff
@@ -491,9 +491,7 @@ where
     }
     macro_rules! f {
         () => {
-            CString::new((*fi.next().unwrap()).clone())
-                .unwrap()
-                .as_ptr()
+            fi.next().unwrap().as_ptr()
```
MeirShpilraien, line 494: Wouldn't we have the same problem as before? the `next` will return the CString which will be dropped right away and you will return its pointer. What am I missing?

MeirShpilraien (commented):
Small question

src/raw.rs
MeirShpilraien: I do not think that the macro is the problem, working with pointer is unsafe and should be done carefully. Regardless, I do not like this macro as well ...

MeirShpilraien: We should probably add a proc macro on redismodule-rs-macros that will generated this call and this way we will also not be limited to the number of fields...

slavak: I did some testing to check that the destructor is called at the correct time, but on second (third? fourth?) thought that testing might have been flawed and you might be right about the lifetime still being wrong.

src/raw.rs
slavak: If the field names can contain internal NULLs then the entire use of `REDISMODULE_HASH_CFIELDS` is wrong. That would arguably be an implementation bug and not an error condition, so panicking is the correct way to handle it. (We can use an `expect` if we want to be nicer.)

A discussion on whether we should support non-string field names might be called for, but is outside the scope of this issue I think.

src/raw.rs
slavak: Not sure I understand how a proc macro would work, but it's been a while since I've dug into the details of Rust macros. I'm all in favor of it if it's possible.

MeirShpilraien: I suggest we will fix the current issue first, and open another issue to define this proc macro.

src/raw.rs
MeirShpilraien: I believe you need to collect the CString to another vector and then iterate over this vector by references (So you will get references to the CString's).

slavak: Yeah I think you might be right. I was trying to handle it without extra allocations.

slavak:
Pushed a new commit that should properly resolve this by (regrettably) allocating a new `Vec`.

Unfortunately I don't have any easy way of testing this actually works. Would appreciate some help on that front.

MeirShpilraien (approved):
LGTM
I was wondering how come our valgrind run did not catch it, but then I notice we do not have such run :) (seems like I already open an [issue](https://github.com/RedisLabsModules/redismodule-rs/issues/331) to add it), so till then I believe we can simply make sure that the warning is gone.
I believe we should also open a followup issue to design this API with proc macros (as discussed here).

PR #355, opened 2023-07-17
Improves the info handler API.

iddm:
The info handler is a function which is invoked in order to handle an "INFO" command call of Redis. This call, when the module name is specified, is redirected to the module's handler function.

This commit brings some user-experience improvements for the handler implementation:

1. A possibility to build a response from a struct using a derive.
2. A possibility to use a `InfoContextBuilder` facility which conveniently wraps the API to enable/disable certain code paths of the Redis API so, that it is possible to use it the correct way only.

### DIFF
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -809,6 +809,28 @@ impl Context {
             .into()
         }
     );
+
+    api!(
+        [RedisModule_AvoidReplicaTraffic],
+        /// Returns true if a client sent the CLIENT PAUSE command to the server or
+        /// if Redis Cluster does a manual failover, pausing the clients.
+        /// This is needed when we have a master with replicas, and want to write,
+        /// without adding further data to the replication channel, that the replicas
+        /// replication offset, match the one of the master. When this happens, it is
+        /// safe to failover the master without data loss.
+        ///
+        /// However modules may generate traffic by calling commands or directly send
+        /// data to the replication stream.
+        ///
+        /// So modules may want to try to avoid very heavy background work that has
+        /// the effect of creating data to the replication channel, when this function
+        /// returns true. This is mostly useful for modules that have background
+        /// garbage collection tasks, or that do writes and replicate such writes
+        /// periodically in timer callbacks or other periodic callbacks.
+        pub fn avoid_replication_traffic(&self) -> bool {
+            unsafe { RedisModule_AvoidReplicaTraffic() == 1 }
+        }
+    );
 }
 
 extern "C" fn post_notification_job_free_callback<F: FnOnce(&Context)>(pd: *mut c_void) {


examples/test_helper.rs
```diff
@@ -31,10 +33,20 @@ fn test_helper_err(ctx: &Context, args: Vec<RedisString>) -> RedisResult {
...
-        ctx.add_info_field_str("field", "test_helper_value");
-    }
+#[derive(Debug, Clone, InfoSection)]
+struct InfoData {
+    field: String,
+    dictionary: HashMap<String, String>,
+}
+
+fn add_info(ctx: &InfoContext, _for_crash_report: bool) -> RedisResult<()> {
+    let mut dictionary = HashMap::new();
```
MeirShpilraien, line 43: Can we keep the exact old behaviour here?

examples/info_handler_macro.rs
```diff
@@ -0,0 +1,67 @@
...
+
+//////////////////////////////////////////////////////
+
+redis_module! {
+    name: "test_helper",
+    version: 1,
+    allocator: (redis_module::alloc::RedisAlloc, redis_module::alloc::RedisAlloc),
+    data_types: [],
+    commands: [
+        ["test_helper.version", test_helper_version, "", 0, 0, 0],
```
MeirShpilraien, line 62: Why do you need all those commands?

redismodule-rs-macros/src/info_section.rs
```diff
@@ -0,0 +1,109 @@
...
+use proc_macro2::Ident;
+use quote::{quote, ToTokens};
+use syn::{parse_macro_input, Data, DataStruct, DeriveInput, Fields};
+
+/// Returns `true` if the `type_string` provided is a type of a map
+/// supported by the [`crate::InfoSection`].
+pub fn is_supported_map(type_string: &str) -> bool {
+    /// A list of supported maps which can be converted to a dictionary for
+    /// the [`redis_module::InfoContext`].
+    const ALL: [&str; 2] = ["BTreeMap", "HashMap"];
```
MeirShpilraien, line 11: Why do we need this validation, it will raise a compilation error anyway if the type will not implement the from trait no?

examples/test_helper.rs
MeirShpilraien: I agree but we should keep it just to make sure it works, maybe even add a deprecation notice on this example?

redismodule-rs-macros/src/info_section.rs
MeirShpilraien: So this means we can not put another struct as a dictionary section?

redismodule-rs-macros/src/lib.rs
```diff
@@ -360,3 +361,92 @@ pub fn cron_event_handler(_attr: TokenStream, item: TokenStream) -> TokenStream
...
+/// }
+/// ```
+///
+/// This procedural macro only implements an easy way to convert objects
+/// of this struct, it doesn't automatically do anything. To actually
+/// make use of this, we must return an object of this struct from the
+/// corresponding handler (`info` handler):
+///
+/// ```rust,no_run,ignore
+/// static mut INFO: Info = Info::new();
```
MeirShpilraien, line 421: Missing the derive ...

MeirShpilraien: Ohh forget it, its on the struct def...

src/context/call_reply.rs
```diff
@@ -822,7 +822,9 @@ impl<'ctx> FutureCallReply<'ctx> {
 
 impl<'ctx> Drop for FutureCallReply<'ctx> {
     fn drop(&mut self) {
-        self.reply.map(|v| free_call_reply(v.as_ptr()));
+        if let Some(v) = self.reply {
+            free_call_reply(v.as_ptr());
+        }
```
MeirShpilraien, line 827: Just curious, why changing this?

src/context/mod.rs
```diff
@@ -849,6 +854,351 @@ bitflags! {
...
+                    self.context.ctx,
+                    Some(section_name),
+                ))?;
+
+                self.add_top_level_fields(section_fields)
+            })
+    }
+
+    /// Sends the info accumulated so far to the [`InfoContext`].
+    pub fn build_info(self) -> RedisResult<&'a InfoContext> {
```
MeirShpilraien, line 1171: If I understand correctly, this will return an error if Redis returns status REDISMODULE_ERR. If so will lose the InfoContext and will not be able to continue add more sections. I believe this is not the expected behaviour, if Redis returns REDISMODULE_ERR it means that the section was not requested and we should not add it, but it does not mean we should stop adding more sections.

MeirShpilraien: Please correct me if I misunderstood the code ...

src/key.rs
```diff
@@ -563,15 +563,6 @@ impl<'a> StringDMA<'a> {
     }
 }
 
-impl From<raw::Status> for Result<(), RedisError> {
```
MeirShpilraien, line 566: Was this moved somewhere?

src/lib.rs
```diff
@@ -58,22 +63,44 @@ pub static MODULE_CONTEXT: DetachedContext = DetachedContext::new();
...
+pub type InfoHandlerFunctionType = fn(&InfoContext, bool) -> RedisResult<()>;
+
+/// Default "INFO" command handler for the module.
+///
+/// This function can be invoked, for example, by sending `INFO modules`
+/// through the RESP protocol.
+pub fn basic_info_command_handler(ctx: &InfoContext, for_crash_report: bool) {
+    if for_crash_report {
+        if let Err(e) = add_trace_info(ctx) {
+            log::error!("Couldn't send info for the module: {e}");
```
MeirShpilraien, line 95: Here also, if the trace section is not requested it will not call the other info handlers.

tests/integration.rs
```diff
@@ -113,19 +113,24 @@ fn test_command_name() -> Result<()> {
...
-    assert!(res.contains("test_helper_field:test_helper_value"));
+    MODULES.iter().try_for_each(|module| {
+        let port: u16 = 6483;
+        let _guards = vec![start_redis_server_with_module(module, port)
+            .with_context(|| "failed to start redis server")?];
+        let mut con =
+            get_redis_connection(port).with_context(|| "failed to connect to redis server")?;
 
-    Ok(())
+        let res: String = redis::cmd("INFO")
```
MeirShpilraien, line 125: Let add more sections and lets also add a test that asks for specific section and make sure it is working.

MeirShpilraien (commented):
Look great :+1: 
The main comment is about not stopping the info report if Redis returns REDISMODULE_ERR (which simply means to skip the section IIRC)

redismodule-rs-macros/src/info_section.rs
MeirShpilraien: Feels like we need another derive, `InfoDictionary`...

examples/test_helper.rs
MeirShpilraien: Sounds good, I only care to keep some sanity test that make sure the old API still works. Even if we are sure it is not broken now, it will keep us from breaking it in the future until we will drop it...

redismodule-rs-macros/src/info_section.rs
MeirShpilraien: And also an attribute on the field that tells if its a dictionary probably... but I do not think it is mandatory for now so we can save it for future improvements ...

examples/test_helper.rs
MeirShpilraien: Sorry for been pedantic but I think its very important. It is still not exactly the same, the return type is changed:
https://github.com/RedisLabsModules/redismodule-rs/pull/355/files#diff-ccfc40870ea0b55ee48ea454932642f33976a4c616ab31982d299902114a28c8R35

redismodule-rs-macros/src/info_section.rs
MeirShpilraien: > And if we mark it with an attribute, we still won't be sure (it still may fail).
It will failed on compile time, but we will know which code to generate for the field ...

MeirShpilraien (approved):
:+1:

MeirShpilraien (approved):

PR #357, opened 2023-07-25
Force post notification job callback to static lifetime

MeirShpilraien:
Post notification job callback is kept by Redis for unknown time and so we must force it live for the static lifetime. Notice that this does not mean that the value must live forever, it just means that it can not hold any none static references. For more information: https://practice.rs/lifetime/static.html#t-static


MeirShpilraien:
No, its needed in general, the API today is wrong. The callback can borrow references to object that will be free before the callback is called.

PR #362, opened 2023-09-13
open key flags

DvirDukhan:
This PR adds the ability to open key for read or write with additional flags
It adds the function `open_with_flags` both for `RedisKeyWritable` and `RedisKey` as well as the following flags:
```
    /// Avoid touching the LRU/LFU of the key when opened.
    NOTOUCH,
    /// Don't trigger keyspace event on key misses.
    NONOTIFY,
    /// Don't update keyspace hits/misses counters
    NOSTATS,
    /// Avoid deleting lazy expired keys.
    NOEXPIRE,
    /// Avoid any effects from fetching the key.
    NOEFFECTS
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
-cargo build --features experimental-api,test --all --all-targets
+cargo build --all --all-targets --no-default-features
 
 
--- a/examples/open_key_with_flags.rs
+++ b/examples/open_key_with_flags.rs
@@ -1,0 +1,79 @@
+use std::{thread, time::Duration};
+
+use redis_module::redisvalue::RedisValueKey;
+use redis_module::{redis_module, Context, RedisError, RedisResult, RedisString, NextArg, RedisValue};
+use redis_module::key::KeyFlag;
+
+
+fn extract_expired_keys_from_stats(stats: RedisValue)->i32{
+    match stats{
+        RedisValue::SimpleString(s) => {
+            s.match_indices("expired_keys:")
+                .map(|(i, _)| i)
+                .last()
+                .map(|i| s[i + 13..i + 14].parse::<i32>().expect("error"))
+                .expect("error")
+        },
+        _ => panic!("expired_keys is not an integer"),
+    }
+}
+
+fn read(ctx: &Context, args: Vec<RedisString>) -> RedisResult {
+    if args.len() < 2 {
+        return Err(RedisError::WrongArity);
+    }
+
+    let mut args = args.into_iter().skip(1);
+    let key_name = args.next_arg()?;
+
+    let key = ctx.open_key_writable(&key_name);
+    let stats = ctx.call("info", &["stats"]).expect("error");
+    key.set_expire(Duration::from_millis(1));
+    thread::sleep(Duration::from_millis(1));
+    let key = ctx.open_key_with_flags(&key_name, &[KeyFlag::NOEFFECTS].to_vec());
+    let stats_after = ctx.call("info", &["stats"]).expect("error");
+    let expired_before = extract_expired_keys_from_stats(stats);
+    let expired_after = extract_expired_keys_from_stats(stats_after);
+    let ret = expired_before == expired_after;
+    match ret {
+        true => Ok(RedisValue::SimpleStringStatic("OK")),
+        false => Err(RedisError::String(format!("stats changed {expired_before:?} {expired_after:?}"))),
+    }
+}
+
+fn write(ctx: &Context, args: Vec<RedisString>) -> RedisResult {
+    if args.len() < 2 {
+        return Err(RedisError::WrongArity);
+    }
+
+    let mut args = args.into_iter().skip(1);
+    let key_name = args.next_arg()?;
+
+    let key = ctx.open_key_writable(&key_name);
+    let stats = ctx.call("info", &["stats"]).expect("error");
+    key.set_expire(Duration::from_millis(1));
+    thread::sleep(Duration::from_millis(1));
+    let key = ctx.open_key_writable_with_flags(&key_name, &[KeyFlag::NOEFFECTS].to_vec());
+    let stats_after = ctx.call("info", &["stats"]).expect("error");
+    
+    let expired_before = extract_expired_keys_from_stats(stats);
+    let expired_after = extract_expired_keys_from_stats(stats_after);
+    let ret = expired_before == expired_after;
+    match ret {
+        true => Ok(RedisValue::SimpleStringStatic("OK")),
+        false => Err(RedisError::String(format!("stats changed {expired_before:?} {expired_after:?}"))),
+    }
+}
+
+//////////////////////////////////////////////////////
+
+redis_module! {
+    name: "open_key_with_flags",
+    version: 1,
+    allocator: (redis_module::alloc::RedisAlloc, redis_module::alloc::RedisAlloc),
+    data_types: [],
+    commands: [
+        ["open_key_with_flags.read", read, "write fast deny-oom", 1, 1, 1],
+        ["open_key_with_flags.write", write, "write fast deny-oom", 1, 1, 1]
+    ],
+}
--- a/examples/open_key_with_flags.rs
+++ b/examples/open_key_with_flags.rs
@@ -21,7 +21,7 @@ fn validate_open_key_with_no_effects(ctx: &Context, key_name: RedisString, read_
     let key = ctx.open_key_writable(&key_name);
     let stats = ctx.call("info", &["stats"])?;
     key.set_expire(Duration::from_millis(1));
-    thread::sleep(Duration::from_millis(1));
+    thread::sleep(Duration::from_millis(2));
     if read_write {
         ctx.open_key_writable_with_flags(&key_name, raw::KeyMode::NOEFFECTS);
     } else {
--- a/examples/open_key_with_flags.rs
+++ b/examples/open_key_with_flags.rs
@@ -1,6 +1,5 @@
 use std::{thread, time::Duration};
 
-use redis_module::redisvalue::RedisValueKey;
 use redis_module::{redis_module, Context, RedisError, RedisResult, RedisString, NextArg, RedisValue};
 use redis_module::key::KeyFlag;
 
@@ -30,7 +29,7 @@ fn read(ctx: &Context, args: Vec<RedisString>) -> RedisResult {
     let stats = ctx.call("info", &["stats"]).expect("error");
     key.set_expire(Duration::from_millis(1));
     thread::sleep(Duration::from_millis(1));
-    let key = ctx.open_key_with_flags(&key_name, &[KeyFlag::NOEFFECTS].to_vec());
+    ctx.open_key_with_flags(&key_name, &[KeyFlag::NOEFFECTS].to_vec());
     let stats_after = ctx.call("info", &["stats"]).expect("error");
     let expired_before = extract_expired_keys_from_stats(stats);
     let expired_after = extract_expired_keys_from_stats(stats_after);
@@ -53,7 +52,7 @@ fn write(ctx: &Context, args: Vec<RedisString>) -> RedisResult {
     let stats = ctx.call("info", &["stats"]).expect("error");
     key.set_expire(Duration::from_millis(1));
     thread::sleep(Duration::from_millis(1));
-    let key = ctx.open_key_writable_with_flags(&key_name, &[KeyFlag::NOEFFECTS].to_vec());
+    ctx.open_key_writable_with_flags(&key_name, &[KeyFlag::NOEFFECTS].to_vec());
     let stats_after = ctx.call("info", &["stats"]).expect("error");
     
     let expired_before = extract_expired_keys_from_stats(stats);
--- a/examples/open_key_with_flags.rs
+++ b/examples/open_key_with_flags.rs
@@ -1,22 +1,57 @@
 use std::{thread, time::Duration};
 
-use redis_module::{redis_module, Context, RedisError, RedisResult, RedisString, NextArg, RedisValue};
-use redis_module::key::KeyFlag;
+use redis_module::{redis_module, Context, RedisError, RedisResult, RedisString, NextArg, RedisValue, raw};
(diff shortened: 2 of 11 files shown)


src/context/mod.rs
```diff
@@ -613,12 +613,23 @@ impl Context {
     pub fn open_key(&self, key: &RedisString) -> RedisKey {
         RedisKey::open(self.ctx, key)
     }
+    
+    #[must_use]
+    pub fn open_key_with_flags(&self, key: &RedisString, flags: &Vec<KeyFlag>) -> RedisKey {
```
MeirShpilraien, line 618: Why not using `bitflags!` for the flags attribute. Currently it is easy to make mistakes such as passing the same flag twice.

MeirShpilraien: Also it is more performant

src/context/mod.rs
```diff
@@ -613,12 +613,23 @@ impl Context {
...
+    }
+
 
     #[must_use]
     pub fn open_key_writable(&self, key: &RedisString) -> RedisKeyWritable {
         RedisKeyWritable::open(self.ctx, key)
     }
 
+    #[must_use]
+    pub fn open_key_writable_with_flags(&self, key: &RedisString, flags: &Vec<KeyFlag>) -> RedisKeyWritable {
```
MeirShpilraien, line 629: Same comment as above

src/key.rs
```diff
@@ -176,6 +196,11 @@ impl RedisKeyWritable {
         Self { ctx, key_inner }
     }
 
+    pub fn open_with_flags(ctx: *mut raw::RedisModuleCtx, key: &RedisString, flags: &Vec<KeyFlag>) -> Self {
+        let key_inner = raw::open_key(ctx, key.inner,  to_raw_mode(KeyMode::ReadWrite) |  to_raw_flags(flags));
```
MeirShpilraien, line 200: If we use `bitflags!` we do not need the `to_raw_flags` transformation

src/key.rs
```diff
@@ -607,6 +632,22 @@ fn to_raw_mode(mode: KeyMode) -> raw::KeyMode {
     }
 }
 
+fn to_raw_flag(flag: KeyFlag)->raw::KeyMode {
```
MeirShpilraien, line 635: We are not gone need it but in general it is probably better to implement the `From` trait

src/raw.rs
```diff
@@ -29,6 +29,17 @@ bitflags! {
     pub struct KeyMode: c_int {
         const READ = REDISMODULE_READ as c_int;
         const WRITE = REDISMODULE_WRITE as c_int;
+
+        /// Avoid touching the LRU/LFU of the key when opened.
```
MeirShpilraien, line 33: We said we will split it to a separate struct no?

examples/open_key_with_flags.rs
```diff
@@ -0,0 +1,78 @@
...
+
+//////////////////////////////////////////////////////
+
+redis_module! {
+    name: "open_key_with_flags",
+    version: 1,
+    allocator: (redis_module::alloc::RedisAlloc, redis_module::alloc::RedisAlloc),
+    data_types: [],
+    commands: [
+        ["open_key_with_flags.read", read, "write fast deny-oom", 1, 1, 1],
```
MeirShpilraien, line 75: Lets use the proc macro instead of defining the command here, we want to drop this syntax. See: https://github.com/RedisLabsModules/redismodule-rs/blob/12ba3dd9d178ff4b793cd11c53937dca293d635d/examples/proc_macro_commands.rs#L34

examples/open_key_with_flags.rs
```diff
@@ -0,0 +1,78 @@
...
+        return Err(RedisError::WrongArity);
+    }
+
+    let mut args = args.into_iter().skip(1);
+    let key_name = args.next_arg()?;
+
+    let key = ctx.open_key_writable(&key_name);
+    let stats = ctx.call("info", &["stats"]).expect("error");
+    key.set_expire(Duration::from_millis(1));
+    thread::sleep(Duration::from_millis(1));
```
MeirShpilraien, line 54: This can be really flaky. You might get the test pass but only because the key was not really expired.

examples/open_key_with_flags.rs
```diff
@@ -0,0 +1,78 @@
...
+fn write(ctx: &Context, args: Vec<RedisString>) -> RedisResult {
+    if args.len() < 2 {
+        return Err(RedisError::WrongArity);
+    }
+
+    let mut args = args.into_iter().skip(1);
+    let key_name = args.next_arg()?;
+
+    let key = ctx.open_key_writable(&key_name);
+    let stats = ctx.call("info", &["stats"]).expect("error");
```
MeirShpilraien, line 52: Why not use `?` and raise the error?

examples/open_key_with_flags.rs
```diff
@@ -0,0 +1,78 @@
...
+    thread::sleep(Duration::from_millis(1));
+    ctx.open_key_writable_with_flags(&key_name, &[KeyFlag::NOEFFECTS].to_vec());
+    let stats_after = ctx.call("info", &["stats"]).expect("error");
+    
+    let expired_before = extract_expired_keys_from_stats(stats);
+    let expired_after = extract_expired_keys_from_stats(stats_after);
+    let ret = expired_before == expired_after;
+    match ret {
+        true => Ok(RedisValue::SimpleStringStatic("OK")),
+        false => Err(RedisError::String(format!("stats changed {expired_before:?} {expired_after:?}"))),
```
MeirShpilraien, line 63:
```suggestion
    if expired_before == expired_after {
        Ok(RedisValue::SimpleStringStatic("OK"))
    } else {
        Err(RedisError::String(format!("stats changed {expired_before:?} {expired_after:?}")))
    }
```

examples/open_key_with_flags.rs
```diff
@@ -0,0 +1,78 @@
...
+    thread::sleep(Duration::from_millis(1));
+    ctx.open_key_with_flags(&key_name, &[KeyFlag::NOEFFECTS].to_vec());
+    let stats_after = ctx.call("info", &["stats"]).expect("error");
+    let expired_before = extract_expired_keys_from_stats(stats);
+    let expired_after = extract_expired_keys_from_stats(stats_after);
+    let ret = expired_before == expired_after;
+    match ret {
+        true => Ok(RedisValue::SimpleStringStatic("OK")),
+        false => Err(RedisError::String(format!("stats changed {expired_before:?} {expired_after:?}"))),
+    }
```
MeirShpilraien, line 40: I have mostly the same comment on this test as the one bellow, maybe we can unified the code somehow?

MeirShpilraien (commented):
Looks good
I would change the test to simply register to expired event and count how many expired we had.
Also I would get the expiration value from outside (from the integration.rs test itself) and make the open_key_with_flags.rs much smaller and simpler.

src/raw.rs
DvirDukhan: Any Ideas how to `|` between other types? Let's slack this one?

examples/open_key_with_flags.rs
DvirDukhan: Increased to 2 milis. Thoughts?

DvirDukhan:
> Looks good I would change the test to simply register to expired event and count how many expired we had. Also I would get the expiration value from outside (from the integration.rs test itself) and make the open_key_with_flags.rs much smaller and simpler.

Don't we want to avoid Redis main thread doing anything while the module is running and open keys? If we count expirations from the outside, we will eventually have expirations.

build.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
-cargo build --features experimental-api,test --all --all-targets
+cargo build --all --all-targets --no-default-features
```
MeirShpilraien, line 2: Why this change?

examples/open_key_with_flags.rs
MeirShpilraien: I honestly would prefer to just expose a function that perform the read/write open of the key and check the stats of the number of expiration from the test itself (remember to disable active expire).

MeirShpilraien:
> > Looks good I would change the test to simply register to expired event and count how many expired we had. Also I would get the expiration value from outside (from the integration.rs test itself) and make the open_key_with_flags.rs much smaller and simpler.
> 
> Don't we want to avoid Redis main thread doing anything while the module is running and open keys? If we count expirations from the outside, we will eventually have expirations.

This is why you need to disable active expire.

examples/open_key_with_flags.rs
```diff
@@ -0,0 +1,99 @@
...
+use redis_module_macros::command;
+
+
+fn extract_expired_keys_from_stats(stats: RedisValue)->Result<i32, String>{
+    match stats{
+        RedisValue::SimpleString(s) => {
+            let val = s.match_indices("expired_keys:")
+                .map(|(i, _)| i)
+                .last()
+                .map(|i| s[i + 13..i + 14].parse::<i32>()).unwrap();
```
MeirShpilraien, line 13: Raise the error

examples/open_key_with_flags.rs
```diff
@@ -0,0 +1,99 @@
...
+            Ok(val.ok().unwrap())
```
MeirShpilraien, line 14: I am not sure what this line does why not just return val?

src/raw.rs
```diff
@@ -29,6 +29,17 @@ bitflags! {
...
+        const NOTOUCH = REDISMODULE_OPEN_KEY_NOTOUCH as c_int;
+        /// Don't trigger keyspace event on key misses.
```
MeirShpilraien, line 35: Didn't we say we will split it? Currently the user can give the `WRITE` flag to `RedisKey`?

test.sh
```diff
@@ -1,2 +1,2 @@
 #!/usr/bin/env sh
-cargo test --all-targets --features test,experimental-api
+cargo test --all --all-targets --no-default-features
```
MeirShpilraien, line 2: Why is this change?

MeirShpilraien (commented):
:+1: 
Few more comments

build.sh
MeirShpilraien: ?

src/key.rs
```diff
@@ -50,6 +68,16 @@ impl RedisKey {
         Self { ctx, key_inner }
     }
 
+    pub fn open_with_flags(
```
MeirShpilraien, line 71:
```suggestion
    pub(crate) fn open_with_flags(
```

src/key.rs
```diff
@@ -176,6 +204,20 @@ impl RedisKeyWritable {
...
```
MeirShpilraien, line 207:
```suggestion
    pub(crate) fn open_with_flags(
```

test.sh
MeirShpilraien: ?

tests/integration.rs
```diff
@@ -646,3 +649,65 @@ fn test_call_blocking() -> Result<()> {
...
+        .with_context(|| "failed to run DEBUG SET-ACTIVE-EXPIRE")?;
+
+    for read_write in [true, false].iter() {
+        redis::cmd("set")
+            .arg(&["x", "1"])
+            .query(&mut con)
+            .with_context(|| "failed to run string.set")?;
+
+        // Set experition time to 1 second.
+        redis::cmd("expire")
```
MeirShpilraien, line 674: Why not use `pexpire` and then sleep for 500ms? Should be more then enough no?

tests/integration.rs
```diff
@@ -646,3 +649,65 @@ fn test_call_blocking() -> Result<()> {
...
+    let mut con =
+        get_redis_connection(port).with_context(|| "failed to connect to redis server")?;
+
+    // Avoid active expriation
+    redis::cmd("DEBUG")
+        .arg(&["SET-ACTIVE-EXPIRE", "0"])
+        .query(&mut con)
+        .with_context(|| "failed to run DEBUG SET-ACTIVE-EXPIRE")?;
+
+    for read_write in [true, false].iter() {
```
MeirShpilraien, line 667:
```suggestion
    for cmd in ["open_key_with_flags.write", "open_key_with_flags.read"].into_iter() {
```

tests/integration.rs
```diff
@@ -646,3 +649,65 @@ fn test_call_blocking() -> Result<()> {
...
+            .with_context(|| "failed to run expire")?;
+
+        // Sleep for 2 seconds, ensure expiration time has passed.
+        thread::sleep(Duration::from_secs(2));
+
+        let cmd = if *read_write {
+            "open_key_with_flags.write"
+        } else {
+            "open_key_with_flags.read"
+        };
```
MeirShpilraien, line 686:
```suggestion

```

MeirShpilraien (commented):
Last tiny comments and we can merge

tests/integration.rs
```diff
@@ -646,3 +649,65 @@ fn test_call_blocking() -> Result<()> {
...
+        let stats: String = redis::cmd("info").arg(&["stats"]).query(&mut con)?;
+
+        // Find the number of expired keys, x,  according to the substring "expired_keys:{x}"
+        let expired_keys = stats
+            .match_indices("expired_keys:")
+            .next()
+            .map(|(i, _)| &stats[i..i + "expired_keys:".len() + 1])
+            .and_then(|s| s.split(':').nth(1))
+            .and_then(|s| s.parse::<i32>().ok())
+            .unwrap_or(-1);
```
oshadmi, line 702: @DvirDukhan Consider using a regex to capture the number of expired keys
(also need to make `utils` module public)
```suggestion
        use redis_module::utils::get_regexp_captures;
        let expired_keys = get_regexp_captures(stats.as_str(), r"\bexpired_keys:([0-9]+)\b")
            .unwrap()[0]
            .parse::<i32>()
            .unwrap_or(-1);
```

build.sh
MeirShpilraien: Match the makefile command

test.sh
MeirShpilraien: Match the makefile command

MeirShpilraien (dismissed):

oshadmi (approved):

PR #375, opened 2023-11-22
Update dependencies  and refactor the FFI calls

vityafx:
This greatly simplifies the way the API calls are done, as well as the resulting type conversion is now hidden, which greatly reduces the number of places for the change in the future.

### DIFF
--- a/src/context/commands.rs
+++ b/src/context/commands.rs
@@ -454,12 +454,12 @@ api! {[
             // the only CString pointers which are not freed are those of the key_specs, lets free them here.
             key_specs.into_iter().for_each(|v|{
                 if !v.notes.is_null() {
-                    unsafe{CString::from_raw(v.notes as *mut c_char)};
+                    unsafe{ drop(CString::from_raw(v.notes as *mut c_char)) };
                 }
                 if v.begin_search_type == raw::RedisModuleKeySpecBeginSearchType_REDISMODULE_KSPEC_BS_KEYWORD {
                     let keyword = unsafe{v.bs.keyword.keyword};
                     if !keyword.is_null() {
-                        unsafe{CString::from_raw(v.bs.keyword.keyword as *mut c_char)};
+                        unsafe{ drop(CString::from_raw(v.bs.keyword.keyword as *mut c_char)) };
                     }
                 }
             });
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -476,8 +476,7 @@ impl Context {
 
     #[allow(clippy::must_use_candidate)]
     pub fn reply_error_string(&self, s: &str) -> raw::Status {
-        let msg = Self::str_as_legal_resp_string(s);
-        unsafe { raw::RedisModule_ReplyWithError.unwrap()(self.ctx, msg.as_ptr()).into() }
+        raw::reply_with_error(self.ctx, s)
     }
 
     pub fn reply_with_key(&self, result: RedisValueKey) -> raw::Status {
@@ -590,14 +589,14 @@ impl Context {
 
             Ok(RedisValue::StaticError(s)) => self.reply_error_string(s),
 
-            Err(RedisError::WrongArity) => unsafe {
+            Err(RedisError::WrongArity) => {
                 if self.is_keys_position_request() {
                     // We can't return a result since we don't have a client
                     raw::Status::Err
                 } else {
-                    raw::RedisModule_WrongArity.unwrap()(self.ctx).into()
+                    raw::wrong_arity(self.ctx)
                 }
-            },
+            }
 
             Err(RedisError::WrongType) => {
                 self.reply_error_string(RedisError::WrongType.to_string().as_str())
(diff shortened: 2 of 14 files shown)


src/raw.rs
```diff
@@ -208,83 +240,110 @@ pub const REDISMODULE_HASH_DELETE: *const RedisModuleString = 1 as *const RedisM
...
+}
+
+macro_rules! generate_transparent_binding {
+    ($(#[$outer:meta])* $vis:vis $rust_name:ident => $raw_function:ident, $ret_type:ty, $(($arg:ident, $typ:ty)),*) => {
+        $(#[$outer])*
+        #[allow(clippy::not_unsafe_ptr_arg_deref)]
+        $vis fn $rust_name($($arg: $typ),*) -> $ret_type {
+            raw_call!($raw_function, $($arg),*)
+        }
+    };
```
vityafx: Discussion: This is one of the ways we could potentially leverage to simplify writing one-to-one function calls (one-liners), which perform some basic processing. I thought this wasn't elegant, and the best way would be to have a procedural macro instead. An example of the invocation:

```rust
generate_transparent_binding!(
    /// # Panics
    ///
    /// Panics if the Redis server doesn't support replying with bool (since RESP3).
    pub call_reply_set_element => RedisModule_CallReplySetElement,
    *mut RedisModuleCallReply,
    (reply, *mut RedisModuleCallReply),
    (idx, usize)
);
```

src/context/commands.rs
```diff
@@ -454,12 +454,12 @@ api! {[
             // the only CString pointers which are not freed are those of the key_specs, lets free them here.
             key_specs.into_iter().for_each(|v|{
                 if !v.notes.is_null() {
-                    unsafe{CString::from_raw(v.notes as *mut c_char)};
+                    unsafe{ drop(CString::from_raw(v.notes as *mut c_char)) };
```
MeirShpilraien, line 457: Why calling the drop inside the unsafe block?

src/raw.rs
```diff
@@ -208,81 +240,88 @@ pub const REDISMODULE_HASH_DELETE: *const RedisModuleString = 1 as *const RedisM
 
 #[allow(clippy::not_unsafe_ptr_arg_deref)]
 pub fn call_reply_type(reply: *mut RedisModuleCallReply) -> ReplyType {
```
MeirShpilraien, line 242: I believe all those functions should be `public(crate)` we do not want the user to use them directly right? I know its a breaking change, maybe this PR should be on the next major version?

src/raw.rs
```diff
@@ -357,103 +396,97 @@ pub(crate) fn open_key_with_flags(
...
+    redis_call!(RedisModule_ReplyWithAttribute(ctx, len))
 }
 
 #[allow(clippy::not_unsafe_ptr_arg_deref)]
-pub fn reply_with_error(ctx: *mut RedisModuleCtx, err: *const c_char) {
-    unsafe {
-        let msg = Context::str_as_legal_resp_string(CStr::from_ptr(err).to_str().unwrap());
-        RedisModule_ReplyWithError.unwrap()(ctx, msg.as_ptr());
-    }
+pub fn reply_with_error<S: AsRef<str>>(ctx: *mut RedisModuleCtx, err: S) -> Status {
```
MeirShpilraien, line 439: This was used to get `*const c_char`, I believe it will not accept it now right? So its a breaking change. Another reason why this should be on the next major release.

MeirShpilraien (commented):
Few small comments. My main concern is that there is a lot of  changes and its hard to spot what we might have missed. This is why I believe this should be part of the next major version. And if we do that, maybe we should already mark the internal functions as `pub(crate)` so they will not be used from outside wrongly?

src/raw.rs
MeirShpilraien: I agree that procedural macro is better here. Can make all this glue code redudent.

src/context/mod.rs
```diff
@@ -781,53 +780,38 @@ impl Context {
...
-        }
-        .into();
+        let acl_permission_result =
+            raw::acl_check_key_permissions(user, key_name.inner, permissions.bits());
         unsafe { raw::RedisModule_FreeModuleUser.unwrap()(user) };
         let acl_permission_result: Result<(), &str> = acl_permission_result.into();
         acl_permission_result.map_err(|_e| RedisError::Str("User does not have permissions on key"))
     }
 
-    api!(
```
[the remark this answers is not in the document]
vityafx: After I introduced the `redis_call` here, the compiler started throwing a warning here due to the `unsafe` used within the `api!` macro.

src/raw.rs
vityafx: I was also thinking about that. I was wondering why those were `pub` and visible for everyone, but didn't dare to change: I thought there was a reason it was so. If you think the same as I and there are no reasons, I suggest we do break the compatibility here and make those `pub(crate)` and only expose the Rust API we control, and not these functions.

src/raw.rs
vityafx: Yes, `*const char`; indeed, it breaks the compatibility. This is also why all of this implementation detail should be hidden behind `pub(crate)` too.

src/raw.rs
MeirShpilraien: I agree, but we can only do it on a major version.

src/context/mod.rs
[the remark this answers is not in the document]
MeirShpilraien: So we lose the functionality of the `api!` macro which depends on the Redis version would have the change function return value.

src/raw.rs
vityafx: Now, it is hidden behind `pub(crate)` (the whole module).

src/raw.rs
vityafx: Now, it is hidden behind `pub(crate)` (the whole module).

src/context/mod.rs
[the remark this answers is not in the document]
vityafx: Reverted it back.

src/context/commands.rs
vityafx: Moved outside.

src/context/commands.rs
```diff
@@ -454,12 +454,12 @@ api! {[
...
                 }
                 if v.begin_search_type == raw::RedisModuleKeySpecBeginSearchType_REDISMODULE_KSPEC_BS_KEYWORD {
                     let keyword = unsafe{v.bs.keyword.keyword};
                     if !keyword.is_null() {
-                        unsafe{CString::from_raw(v.bs.keyword.keyword as *mut c_char)};
+                        unsafe{ drop(CString::from_raw(v.bs.keyword.keyword as *mut c_char)) };
```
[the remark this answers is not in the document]
vityafx: Moved outside.

src/raw.rs
```diff
@@ -1,28 +1,16 @@
...
 use enum_primitive_derive::Primitive;
 use libc::size_t;
 
 use crate::error::Error;
 pub use crate::redisraw::bindings::*;
-use crate::{context::StrCallArgs, RedisString};
-use crate::{RedisBuffer, RedisError};
-
-const GENERIC_ERROR_MESSAGE: &str = "Generic error.";
+use crate::{context::StrCallArgs, RedisBuffer, RedisError, RedisString, Status};
```
MeirShpilraien, line 13: Do we need to change all those functions on this file to `public(crate)` or do we count on the fact that we are not exposing this mod as public?

MeirShpilraien: I guess this is the answer: https://github.com/RedisLabsModules/redismodule-rs/pull/375#discussion_r1404184390

MeirShpilraien (approved):
:+1: 
Let make sure this change is only added to the next major version as it contains breaking changes.

src/raw.rs
vityafx: Yes :-)

PR #376, opened 2023-11-24
Simplify the integration test code

iddm:


tests/integration.rs
```diff
@@ -1,21 +1,58 @@
...
+    Ok(vec![start_redis_server_with_module(module_name, port)
+        .map_err(|_| "failed to start redis server")?])
+}
+
+struct TestConnection {
+    _guards: Vec<ChildGuard>,
+    connection: Connection,
+}
+
+static mut TEST_PORT: AtomicU16 = AtomicU16::new(6479);
```
MeirShpilraien, line 24: Let use lazy static and avoid the unsafe?

iddm:
@MeirShpilraien, would you please review once again?

ashtul:
@iddm @MeirShpilraien 
Why not move the change into `utils.rs` and expose it as part of the crate?
It would allow users to use the code to test their modules.

iddm:
> @iddm @MeirShpilraien Why not move the change into `utils.rs` and expose it as part of the crate? It would allow users to use the code to test their modules.

We need it just in one single file, there is no need to create a file of 20 lines of code to just fully import it later from within another file. Doesn't really contribute to the cleanliness of the code here. If at least we had two files for that, - sure.

tests/integration.rs
iddm: I am removing the `mut` and the `unsafe`, so this remains a piece of code without `lazy_static`.

MeirShpilraien (approved):

MeirShpilraien:
> We need it just in one single file, there is no need to create a file of 20 lines of code to just fully import it later from within another file. Doesn't really contribute to the cleanliness of the code here. If at least we had two files for that, - sure.

I agree but also I think that its a nice idea that users will be able to test their module with rust and run it with `cargo test`. I believe that if we want to do it, we should make a crate that will expose all the needed functionality (setting up Redis server with the module and with/without replication or on cluster  mode, aof enabled/disabled, and probably there are much more things that we want to do). That said, all this functionalities are already exists on RLTest so maybe its not worth it?

iddm:
@MeirShpilraien can you just merge it?

tests/integration.rs
```diff
@@ -461,69 +391,67 @@ fn test_server_event() -> Result<()> {
...
 fn test_configuration() -> Result<()> {
-    let port: u16 = 6495;
-    let _guards = vec![start_redis_server_with_module("configuration", port)
-        .with_context(|| "failed to start redis server")?];
+    let mut con = TestConnection::new("configuration");
 
-    let config_get = |config: &str| -> Result<String> {
-        let mut con =
-            get_redis_connection(port).with_context(|| "failed to connect to redis server")?;
+    let config_get = |con: &mut TestConnection, config: &str| -> Result<String> {
```
MeirShpilraien, line 396: I wonder why the closure itself can not borrow the con here (as it was before)?

MeirShpilraien (approved):

PR #377, opened 2023-12-03
Add NEW key space notification flag

vityafx:


src/raw.rs
```diff
@@ -134,9 +134,15 @@ bitflags! {
         const EXPIRED = REDISMODULE_NOTIFY_EXPIRED;
         const EVICTED = REDISMODULE_NOTIFY_EVICTED;
         const STREAM = REDISMODULE_NOTIFY_STREAM;
+        #[cfg(any(
+            feature = "min-redis-compatibility-version-7-0",
+            feature = "min-redis-compatibility-version-7-2"
+        ))]
+        const NEW = REDISMODULE_NOTIFY_NEW;
```
MeirShpilraien, line 141: What if a module which want to be comparable with redis 6 but still want to register to this NEW notification in case it is running on redis 7? Can we think how we can still support this case?

src/raw.rs
```diff
@@ -134,9 +134,15 @@ bitflags! {
...
         const MODULE = REDISMODULE_NOTIFY_MODULE;
         const LOADED = REDISMODULE_NOTIFY_LOADED;
         const MISSED = REDISMODULE_NOTIFY_KEY_MISS;
+        /// Does not include the [`Self::MISSED`].
```
MeirShpilraien, line 145: Its also not include the NEW right?

src/raw.rs
MeirShpilraien: I did not mean that the module want to use this flag on Redis 6. I meant that the module want to be Redis 6 compatible and conditionally gets the NEW notification in case it runs on Redis >= 7.0.1.

vityafx: Then we can just remove the `#[cfg]`, I guess. Redis should check the flags too, if the module doesn't do that.

src/raw.rs
vityafx: Please check the rustdoc again.

src/macros.rs
```diff
@@ -68,6 +68,16 @@ macro_rules! redis_event_handler {
...
 
+        let all_available_flags = $crate::raw::get_keyspace_notification_flags_all();
+        if !all_available_flags.contains($event_type) {
+            let not_supported = $event_type.difference(all_available_flags);
+            log::warn!("The event notification flags set aren't supported: {not_supported:?}");
+
+            // $crate::Context::new($ctx).log_warning(&format!(
+            //     "The event notification flags set aren't supported: {not_supported:?}"
+            // ));
+        }
```
MeirShpilraien, line 79: We can use an else here no? Only register to the event if it is supported. This way we do not count of Redis to ignore it and not return an error. IIRC the `RedisModule_SubscribeToKeyspaceEvents` does return an success/failed indication so Redis might one day decide to validate this input.

MeirShpilraien (changes requested):
Last comment

deny.toml
```diff
@@ -281,6 +281,10 @@ include-dependencies = true
 name = "redis-module-macros"
 allow-globs = ["*"]
 
+[[bans.build.bypass]]
+name = "redis-module-macros-internals"
+allow-globs = ["*"]
```
MeirShpilraien, line 286: How come it was not needed before?

MeirShpilraien (approved):
:+1:

deny.toml
vityafx: I don't know. But there are only build scripts so far. I think we should patch cargo-deny to be able to exclude build scripts from the binary detection. This has also been the case a few weeks ago with v8-rs where I suggested not creating the .so files too.

MeirShpilraien: But I did not see there was any build failures. So I did you sudenly concluded it is needed?

vityafx: Because I see those locally.

PR #386, opened 2024-07-25
Derive hash and comparison traits for RedisAlloc

iddm:


iddm:
CircleCI doesn't support macos anymore. I suggest we just merge and fix the CI later.

src/alloc.rs
```diff
@@ -22,7 +22,7 @@ const REDIS_ALLOCATOR_NOT_AVAILABLE_MESSAGE: &str =
 /// Defines the Redis allocator. This allocator delegates the allocation
 /// and deallocation tasks to the Redis server when available, otherwise
 /// it panics.
-#[derive(Default, Debug, Copy, Clone)]
+#[derive(Default, Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
MeirShpilraien, line 25: Why do you need to compare the allocator? Or hash it?

iddm: I don't need it right here, but if the struct wants to wrap it within another struct and do some counting and simply wants to derive a trait, not having those for `RedisAlloc` disallows that and requires implementing it manually. This is more just for convenience.

MeirShpilraien (approved):

PR #387, opened 2024-09-03
Added support for defrag API.

MeirShpilraien:
The PR adds support for defrag API.
The PR introduce a new struct, `DefragContext`, which provide a safe API over `*mut raw::RedisModuleDefragCtx`.

**Notice** that we do expose an unsafe API to create `DefragContext` from `*mut raw::RedisModuleDefragCtx`. This is because we do not have a safe API for datatype registration. User must register an unsafe function as the defrag callback of the datatype and create the `DefragContext` from `*mut raw::RedisModuleDefragCtx`. We should consider adding a safe API for datatype creation.

In addition, the PR introduce 3 new proc macros:

* defrag_function - allows to register a function to defrag global data
* defrag_start_function - allows to register a function to defrag global data when defrag cycle starts.
* defrag_end_function - allows to register a function to defrag global data when defrag cycle ends.

Example and test for the new API were added.

**Notice**: The start and the end callbacks for defrag was added on [this](https://github.com/redis/redis/pull/13509) PR

### DIFF
--- a/examples/data_type.rs
+++ b/examples/data_type.rs
@@ -127,7 +127,7 @@ fn alloc_defragstats(ctx: &Context, _args: Vec<RedisString>) -> RedisResult {
     let num_defrag_start = NUM_DEFRAG_START.lock(ctx);
     let num_defrag_end = NUM_DEFRAG_END.lock(ctx);
     Ok(RedisValue::OrderedMap(
-        vec![
+        [
             (
                 RedisValueKey::String("num_keys_defrag".to_owned()),
                 RedisValue::Integer(*num_keys_defrag as i64),
--- a/examples/data_type.rs
+++ b/examples/data_type.rs
@@ -1,10 +1,24 @@
+use lazy_static::lazy_static;
+use libc::c_int;
+use redis_module::defrag::DefragContext;
 use redis_module::native_types::RedisType;
-use redis_module::{raw, redis_module, Context, NextArg, RedisResult, RedisString};
+use redis_module::redisvalue::RedisValueKey;
+use redis_module::{
+    raw, redis_module, Context, NextArg, RedisGILGuard, RedisResult, RedisString, RedisValue,
+};
+use redis_module_macros::{defrag_end_function, defrag_function, defrag_start_function};
 use std::os::raw::c_void;
 
 #[derive(Debug)]
 struct MyType {
     data: String,
+}
+
+lazy_static! {
+    static ref NUM_KEYS_DEFRAG: RedisGILGuard<usize> = RedisGILGuard::default();
+    static ref NUM_DEFRAG_START: RedisGILGuard<usize> = RedisGILGuard::default();
+    static ref NUM_DEFRAG_END: RedisGILGuard<usize> = RedisGILGuard::default();
+    static ref NUM_DEFRAG_GLOBALS: RedisGILGuard<usize> = RedisGILGuard::default();
 }
 
 static MY_REDIS_TYPE: RedisType = RedisType::new(
@@ -30,7 +44,7 @@ static MY_REDIS_TYPE: RedisType = RedisType::new(
         free_effort: None,
         unlink: None,
         copy: None,
-        defrag: None,
+        defrag: Some(defrag),
 
         copy2: None,
         free_effort2: None,
@@ -41,6 +55,35 @@ static MY_REDIS_TYPE: RedisType = RedisType::new(
 
 unsafe extern "C" fn free(value: *mut c_void) {
     drop(Box::from_raw(value.cast::<MyType>()));
+}
+
+unsafe extern "C" fn defrag(
+    ctx: *mut raw::RedisModuleDefragCtx,
+    _key: *mut raw::RedisModuleString,
+    _value: *mut *mut c_void,
+) -> c_int {
+    let defrag_ctx = DefragContext::new(ctx);
+    let mut num_keys_defrag = NUM_KEYS_DEFRAG.lock(&defrag_ctx);
+    *num_keys_defrag += 1;
+    0
+}
+
+#[defrag_start_function]
+fn defrag_end(defrag_ctx: &DefragContext) {
+    let mut num_defrag_end = NUM_DEFRAG_END.lock(defrag_ctx);
+    *num_defrag_end += 1;
+}
+
+#[defrag_end_function]
+fn defrag_start(defrag_ctx: &DefragContext) {
+    let mut num_defrag_start = NUM_DEFRAG_START.lock(defrag_ctx);
+    *num_defrag_start += 1;
+}
+
+#[defrag_function]
+fn defrag_globals(defrag_ctx: &DefragContext) {
+    let mut num_defrag_globals = NUM_DEFRAG_GLOBALS.lock(defrag_ctx);
+    *num_defrag_globals += 1;
 }
 
 fn alloc_set(ctx: &Context, args: Vec<RedisString>) -> RedisResult {
@@ -78,6 +121,35 @@ fn alloc_get(ctx: &Context, args: Vec<RedisString>) -> RedisResult {
     Ok(value)
 }
 
+fn alloc_defragstats(ctx: &Context, _args: Vec<RedisString>) -> RedisResult {
+    let num_keys_defrag = NUM_KEYS_DEFRAG.lock(ctx);
+    let num_defrag_globals = NUM_DEFRAG_GLOBALS.lock(ctx);
+    let num_defrag_start = NUM_DEFRAG_START.lock(ctx);
+    let num_defrag_end = NUM_DEFRAG_END.lock(ctx);
+    Ok(RedisValue::OrderedMap(
+        vec![
+            (
+                RedisValueKey::String("num_keys_defrag".to_owned()),
+                RedisValue::Integer(*num_keys_defrag as i64),
+            ),
+            (
+                RedisValueKey::String("num_defrag_globals".to_owned()),
+                RedisValue::Integer(*num_defrag_globals as i64),
+            ),
+            (
+                RedisValueKey::String("num_defrag_start".to_owned()),
+                RedisValue::Integer(*num_defrag_start as i64),
+            ),
+            (
+                RedisValueKey::String("num_defrag_end".to_owned()),
+                RedisValue::Integer(*num_defrag_end as i64),
+            ),
+        ]
+        .into_iter()
+        .collect(),
+    ))
+}
+
 //////////////////////////////////////////////////////
 
 redis_module! {
@@ -90,5 +162,6 @@ redis_module! {
     commands: [
         ["alloc.set", alloc_set, "write", 1, 1, 1],
         ["alloc.get", alloc_get, "readonly", 1, 1, 1],
+        ["alloc.defragstats", alloc_defragstats, "readonly", 0, 0, 0]
     ],
 }
--- a/redismodule-rs-macros/src/lib.rs
+++ b/redismodule-rs-macros/src/lib.rs
@@ -451,3 +451,72 @@ pub fn info_command_handler(_attr: TokenStream, item: TokenStream) -> TokenStrea
 pub fn info_section(item: TokenStream) -> TokenStream {
     info_section::info_section(item)
 }
+
+/// Proc macro which is set on a function that need to be called whenever server performs defrag.
+/// The function must accept a [&DefragContext]. If defrag is not supported by the Redis version
+/// the function will never be called.
+///
+/// Example:
+///
+/// ```rust,no_run,ignore
+/// #[defrag_function]
+/// fn defrag(ctx: &DefragContext) { ... }
+/// ```
+#[proc_macro_attribute]
+pub fn defrag_function(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let ast: ItemFn = match syn::parse(item) {
+        Ok(res) => res,
+        Err(e) => return e.to_compile_error().into(),
+    };
+    let gen = quote! {
+        #[linkme::distributed_slice(redis_module::defrag::DEFRAG_FUNCTIONS_LIST)]
+        #ast
+    };
+    gen.into()
+}
+
+/// Proc macro which is set on a function that need to be called whenever server start performs defrag.
+/// The function must accept a [&DefragContext]. If defrag start event is not supported by the Redis version
+/// the function will never be called.
+///
+/// Example:
+///
+/// ```rust,no_run,ignore
+/// #[defrag_start_function]
+/// fn defrag_start(ctx: &DefragContext) { ... }
+/// ```
+#[proc_macro_attribute]
+pub fn defrag_start_function(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let ast: ItemFn = match syn::parse(item) {
+        Ok(res) => res,
+        Err(e) => return e.to_compile_error().into(),
+    };
+    let gen = quote! {
+        #[linkme::distributed_slice(redis_module::defrag::DEFRAG_START_FUNCTIONS_LIST)]
+        #ast
+    };
+    gen.into()
+}
+
+/// Proc macro which is set on a function that need to be called whenever server end performs defrag.
+/// The function must accept a [&DefragContext]. If defrag end event is not supported by the Redis version
+/// the function will never be called.
+///
+/// Example:
+///
+/// ```rust,no_run,ignore
+/// #[defrag_end_function]
+/// fn defrag_end(ctx: &DefragContext) { ... }
+/// ```
+#[proc_macro_attribute]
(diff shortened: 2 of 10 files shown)


redismodule-rs-macros/src/lib.rs
```diff
@@ -451,3 +451,72 @@ pub fn info_command_handler(_attr: TokenStream, item: TokenStream) -> TokenStrea
...
+/// ```rust,no_run,ignore
+/// #[defrag_function]
+/// fn defrag(ctx: &DefragContext) { ... }
+/// ```
+#[proc_macro_attribute]
+pub fn defrag_function(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let ast: ItemFn = match syn::parse(item) {
+        Ok(res) => res,
+        Err(e) => return e.to_compile_error().into(),
+    };
```
ephraimfeldblum, line 470:
```suggestion
    let ast: ItemFn = syn::parse(item).unwrap_or_else(|e| e.to_compile_error().into());
```

redismodule-rs-macros/src/lib.rs
```diff
@@ -451,3 +451,72 @@ pub fn info_command_handler(_attr: TokenStream, item: TokenStream) -> TokenStrea
...
+/// ```rust,no_run,ignore
+/// #[defrag_start_function]
+/// fn defrag_start(ctx: &DefragContext) { ... }
+/// ```
+#[proc_macro_attribute]
+pub fn defrag_start_function(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let ast: ItemFn = match syn::parse(item) {
+        Ok(res) => res,
+        Err(e) => return e.to_compile_error().into(),
+    };
```
ephraimfeldblum, line 493:
```suggestion
    let ast: ItemFn = syn::parse(item).unwrap_or_else(|e| e.to_compile_error().into());
```

redismodule-rs-macros/src/lib.rs
```diff
@@ -451,3 +451,72 @@ pub fn info_command_handler(_attr: TokenStream, item: TokenStream) -> TokenStrea
...
+/// ```rust,no_run,ignore
+/// #[defrag_end_function]
+/// fn defrag_end(ctx: &DefragContext) { ... }
+/// ```
+#[proc_macro_attribute]
+pub fn defrag_end_function(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let ast: ItemFn = match syn::parse(item) {
+        Ok(res) => res,
+        Err(e) => return e.to_compile_error().into(),
+    };
```
ephraimfeldblum, line 516:
```suggestion
    let ast: ItemFn = syn::parse(item).unwrap_or_else(|e| e.to_compile_error().into());
```

src/context/defrag.rs
```diff
@@ -0,0 +1,226 @@
...
+    /// position so it can later use [`Self::get_cursor`] to resume defragging.
+    ///
+    /// When stopped and more work is left to be done, the callback should
+    /// return 1. Otherwise, it should return 0.
+    ///
+    /// NOTE: Modules should consider the frequency in which this function is called,
+    /// so it generally makes sense to do small batches of work in between calls.
+    pub fn should_stop(&self) -> bool {
+        let should_stop = unsafe {
+            RedisModule_DefragShouldStop.expect("RedisModule_DefragShouldStop is NULL")(
```
ephraimfeldblum, line 48: `RedisModule_DefragShouldStop.expect("RedisModule_DefragShouldStop is available")`

According to the Rust docs,
> Recommended Message Style
> We recommend that expect messages are used to describe the reason you expect the Option should be Some.

src/context/defrag.rs
```diff
@@ -0,0 +1,226 @@
...
+    ///
+    /// The cursor may be used by the module to represent some progress into the
+    /// module's data type. Modules may also store additional cursor-related
+    /// information locally and use the cursor as a flag that indicates when
+    /// traversal of a new key begins. This is possible because the API makes
+    /// a guarantee that concurrent defragmentation of multiple keys will
+    /// not be performed.
+    pub fn set_cursor(&self, cursor: u64) -> Status {
+        unsafe {
+            RedisModule_DefragCursorSet.expect("RedisModule_DefragCursorSet is NULL")(
```
ephraimfeldblum, line 78: `RedisModule_DefragCursorSet.expect("RedisModule_DefragCursorSet is available")`

src/context/defrag.rs
```diff
@@ -0,0 +1,226 @@
...
+        }
+        .into()
+    }
+
+    /// Fetch a cursor value that has been previously stored using [`Self::set_cursor`].
+    /// If not called for a late defrag operation, [`Err`] will be returned.
+    pub fn get_cursor(&self) -> Result<u64, RedisError> {
+        let mut cursor: u64 = 0;
+        let res: Status = unsafe {
+            RedisModule_DefragCursorGet.expect("RedisModule_DefragCursorGet is NULL")(
```
ephraimfeldblum, line 91: `RedisModule_DefragCursorGet.expect("RedisModule_DefragCursorGet is available")`

src/context/defrag.rs
```diff
@@ -0,0 +1,226 @@
...
+    /// no other effect.
+    ///
+    /// If a non-NULL value is returned, the caller should use the new pointer instead
+    /// of the old one and update any reference to the old pointer, which must not
+    /// be used again.
+    ///
+    /// The function is unsafe because it is assumed that the pointer is valid and previusly
+    /// allocated. It is considered undefined if this is not the case.
+    pub unsafe fn defrag_realloc<T>(&self, mut ptr: *mut T) -> *mut T {
+        let new_ptr: *mut T = RedisModule_DefragAlloc.expect("RedisModule_DefragAlloc is NULL")(
```
ephraimfeldblum, line 118: `RedisModule_DefragAlloc.expect("RedisModule_DefragAlloc is available")`

src/context/defrag.rs
```diff
@@ -0,0 +1,226 @@
...
+
+    /// Defrag a [RedisString]
+    ///
+    /// NOTE: It is only possible to defrag strings that have a single reference.
+    /// Typically this means strings that was copy/cloned using [RedisString::safe_clone]
+    /// or created using [RedisString::new] will not be defrag and will be returned as is.
+    pub fn defrag_redis_string(&self, mut s: RedisString) -> RedisString {
+        let new_inner = unsafe {
+            RedisModule_DefragRedisModuleString
+                .expect("RedisModule_DefragRedisModuleString is NULL")(
```
ephraimfeldblum, line 149: `RedisModule_DefragRedisModuleString.expect("RedisModule_DefragRedisModuleString is available")`

tests/integration.rs
```diff
@@ -743,3 +745,64 @@ fn test_expire() -> Result<()> {
...
+
+    redis::cmd("config")
+        .arg(&["set", "active-defrag-cycle-min", "99"])
+        .query(&mut con)
+        .with_context(|| "failed to run 'config set active-defrag-cycle-min 99'")?;
+
+    // enable active defrag
+    if let Err(_) = redis::cmd("config")
+        .arg(&["set", "activedefrag", "yes"])
+        .query::<String>(&mut con)
```
ephraimfeldblum, line 781:
```suggestion
    if (...).is_err()
```

ephraimfeldblum (approved):

redismodule-rs-macros/src/lib.rs
MeirShpilraien: OK, I am taking it back, this change is not right as it is not return the `e.to_compile_error().into()`.

redismodule-rs-macros/src/lib.rs
MeirShpilraien: Same comment a above

redismodule-rs-macros/src/lib.rs
MeirShpilraien: Same comment as above

ephraimfeldblum (approved):

tests/integration.rs
```diff
@@ -743,3 +745,65 @@ fn test_expire() -> Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn test_defrag() -> Result<()> {
+    let port: u16 = 6503;
+    let _guards = vec![start_redis_server_with_module("data_type", port)
```
iddm, line 752: So this is just one guard, not a vector of multiple?

tests/integration.rs
```diff
@@ -743,3 +745,65 @@ fn test_expire() -> Result<()> {
...
+        // We are looking at num_defrag_globals because this is supported by all Redis versions
+        // that supports defrag.
+        if *num_defrag_globals > 2 {
+            break;
+        }
+        let duration = SystemTime::now().duration_since(start)?;
+        if duration > Duration::from_secs(30) {
+            return Err(anyhow::Error::msg("Failed waiting for defrag cycle"));
+        }
+    }
```
iddm, line 806: Just to clarify the test case. We are testing here that the process works fine and that we have the two defrag cycles. Is it possible for the tests to ever reach (in practice of in theory) the code which checks for the timeout of `30 seconds`? I wonder when and in what cases this might happen. How can we reduce the possible flakiness of this test?

src/context/defrag.rs
```diff
@@ -0,0 +1,227 @@
...
+        };
+        if !new_inner.is_null() {
+            s.inner = new_inner;
+        }
+        s
+    }
+}
+
+#[distributed_slice()]
+pub static DEFRAG_FUNCTIONS_LIST: [fn(&DefragContext)] = [..];
```
iddm, line 162: I'm just curious: do these need to be public?

examples/data_type.rs
```diff
@@ -43,6 +57,35 @@ unsafe extern "C" fn free(value: *mut c_void) {
     drop(Box::from_raw(value.cast::<MyType>()));
 }
 
+unsafe extern "C" fn defrag(
+    ctx: *mut raw::RedisModuleDefragCtx,
+    _key: *mut raw::RedisModuleString,
+    _value: *mut *mut c_void,
+) -> c_int {
+    let defrag_ctx = DefragContext::new(ctx);
+    let mut num_keys_defrag = NUM_KEYS_DEFRAG.lock(&defrag_ctx);
```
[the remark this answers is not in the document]
iddm: Also, why is the `value` provided to this function? What can be done about it or what is assumed about the implementation to be able to be done about it?

examples/data_type.rs
```diff
@@ -43,6 +57,35 @@ unsafe extern "C" fn free(value: *mut c_void) {
     drop(Box::from_raw(value.cast::<MyType>()));
 }
 
+unsafe extern "C" fn defrag(
+    ctx: *mut raw::RedisModuleDefragCtx,
+    _key: *mut raw::RedisModuleString,
```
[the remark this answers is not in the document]
MeirShpilraien: It can be called more than once and in this case the key will be defrag more than once as expected...

examples/data_type.rs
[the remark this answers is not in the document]
MeirShpilraien: Rust do not provide mutable access to statics. The RedisGILGuard is an indication that the static variable is protected by the Redis lock and can only be access when the Redis lock is help by providing a lock indicator (context).

examples/data_type.rs
```diff
@@ -1,12 +1,26 @@
...
 #[derive(Debug)]
 struct MyType {
     data: String,
 }
 
+lazy_static! {
+    static ref NUM_KEYS_DEFRAG: RedisGILGuard<usize> = RedisGILGuard::default();
+    static ref NUM_DEFRAG_START: RedisGILGuard<usize> = RedisGILGuard::default();
+    static ref NUM_DEFRAG_END: RedisGILGuard<usize> = RedisGILGuard::default();
+    static ref NUM_DEFRAG_GLOBALS: RedisGILGuard<usize> = RedisGILGuard::default();
```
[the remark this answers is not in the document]
iddm: Oh no! I have just noticed that this is an example file :-) Thanks.

examples/data_type.rs
```diff
@@ -78,6 +121,35 @@ fn alloc_get(ctx: &Context, args: Vec<RedisString>) -> RedisResult {
...
+            (
+                RedisValueKey::String("num_defrag_end".to_owned()),
+                RedisValue::Integer(*num_defrag_end as i64),
+            ),
+        ]
+        .into_iter()
+        .collect(),
+    ))
+}
+
```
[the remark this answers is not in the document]
MeirShpilraien: Its an example and also it is probably good enough for any realistic usecase. RESP do not expose a usize result type so this is the best we have.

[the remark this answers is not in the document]
iddm: Yes, but still, a good question for somebody who would like to implement it in his module: is showing negative (overflown values) considered good enough, or should we stick to the maximum number and not overflow, or overflow to `0`?

src/context/defrag.rs
```diff
@@ -0,0 +1,227 @@
...
+        if !new_ptr.is_null() {
+            ptr = new_ptr;
+        }
+        ptr
+    }
+
+    /// Allocate memory using defrag allocator if supported by the
+    /// current Redis server, fallback to regular allocation otherwise.
+    pub fn defrag_alloc<T>(&self, layout: Layout) -> *mut T {
+        unsafe { std::alloc::alloc(layout) }.cast()
```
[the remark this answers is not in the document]
MeirShpilraien: From docs:
```
This function forwards calls to the [GlobalAlloc::alloc](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) method of the allocator registered with the #[global_allocator] attribute if there is one, or the std crate’s default.
```
Global allocator is set on the scope of the entire ELF. This means it will call the global allocator as expected.

src/context/defrag.rs
```diff
@@ -0,0 +1,227 @@
...
+}
+
+extern "C" fn defrag_end_function(defrag_ctx: *mut raw::RedisModuleDefragCtx) {
+    let mut ctx = DefragContext { defrag_ctx };
+    DEFRAG_END_FUNCTIONS_LIST.iter().for_each(|callback| {
+        callback(&mut ctx);
+    });
+}
+
+pub fn register_defrag_functions(ctx: &Context) -> Result<(), RedisError> {
```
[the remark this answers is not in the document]
MeirShpilraien: Its public because it is use from the macro, but sure will add a comment.

tests/integration.rs
MeirShpilraien: I actually not sure why it is a vec, but this is how it is done on all the other tests, will leave to change it on another PR.

tests/integration.rs
MeirShpilraien: > Is it possible for the tests to ever reach (in practice of in theory) the code which checks for the timeout of 30 seconds?

Yes, if we have a bug (or a bug in Redis), the defrag functions will not be called and we will reach a timeout.

> How can we reduce the possible flakiness of this test?

I configured the degrag to be very intensive so 30 seconds is much more than what is actually needed and if we reach it we must have a bug.

src/context/defrag.rs
MeirShpilraien: Yes, used from outside of the crate by out proc macros.

examples/data_type.rs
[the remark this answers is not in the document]
MeirShpilraien: So as I wrote, even if defrag is called 1000 times a second (which it should not), it will overflow in around 296533310 years. I believe it is safe to assume it will not overflow...

src/context/defrag.rs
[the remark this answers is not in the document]
iddm: Thanks.

tests/integration.rs
iddm: Thanks.

src/context/defrag.rs
iddm: Thanks.

src/context/defrag.rs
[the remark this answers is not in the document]
iddm: Thanks!

tests/integration.rs
iddm: No problem.

src/context/defrag.rs
```diff
@@ -0,0 +1,227 @@
...
+
+use crate::{
+    raw, Context, RedisModule_DefragAlloc, RedisModule_DefragCursorGet,
+    RedisModule_DefragCursorSet, RedisModule_DefragRedisModuleString, RedisModule_DefragShouldStop,
+    RedisString, Status,
+};
+use crate::{RedisError, RedisLockIndicator};
+use linkme::distributed_slice;
+
+pub struct DefragContext {
```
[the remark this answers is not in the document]
iddm: I see you also changed it to work with `NonNull` instead, thanks!

YaacovHazan (approved):

PR #389, opened 2024-10-01
Move to GitHub actions

iddm:


iddm:
I don't know why the `macos-latest` or `macos-14` runners aren't picking up the macos jobs, maybe today is a bad day?

The tests currently use Redis 7.2 for all the features. I think the best way is actually to use the versions equal to the features, but those fail.

ephraimfeldblum (approved):
LGTM. good job with removing the makefile and readies.

PR #394, opened 2024-10-31
MOD-8036 move `AddACLCategory` to `redismodule-rs`

ephraimfeldblum:


src/macros.rs
```diff
@@ -49,12 +49,13 @@ macro_rules! redis_command {
                 return $crate::raw::Status::Err as c_int;
             }
 
-            if unsafe {
-                $crate::raw::RedisModule_SetCommandACLCategories.unwrap()(
-                    command,
-                    acl_categories.as_ptr(),
-                )
-            } == $crate::raw::Status::Err as c_int
+            if $crate::raw::RedisModule_SetCommandACLCategories.is_some()
```
iddm, line 52: Looks like a candidate of the if let some and else?

ephraimfeldblum: sure. however that necessitates an `if` within an `if` as the resulting `Some` value will not be usable within the condition of the `if` it was generated from.

eranhd (approved):

eranhd (approved):

PR #395, opened 2024-11-03
MOD-8036 Add tests for 393 and 394

ephraimfeldblum:
- Fixes the problem of naming the meta-variable class `$acl_category` identically to its identifier.
- ~Reduces code duplication on the module side (eg `RedisJSON`), all commands from the module are given the ACL category of the module without having to be expressly listed per command.~
(This is not guaranteed to be correct behavior for all modules. removed).
- Adds a test on the `redismodule-rs` side to ensure it works as expected.
- Allows for a combination of optional ACL categories and mandatory ACL categories to allow for the case of optionally adding categories to commands depending on the state of the Redis Server being accessed (ie v7.2 will fail to add a new ACL category, and therefore any commands that should be linked to that category should be optional). Failing to add optional categories results in a warning. Failing to add mandatory categories is an error.

#393 
#394

### DIFF
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -926,6 +926,113 @@ bitflags! {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub enum AclCategory {
+    #[default]
+    None,
+    Keyspace,
+    Read,
+    Write,
+    Set,
+    SortedSet,
+    List,
+    Hash,
+    String,
+    Bitmap,
+    HyperLogLog,
+    Geo,
+    Stream,
+    PubSub,
+    Admin,
+    Fast,
+    Slow,
+    Blocking,
+    Dangerous,
+    Connection,
+    Transaction,
+    Scripting,
+    Single(String),
+    Multi(Vec<AclCategory>),
+}
+
+impl From<Vec<AclCategory>> for AclCategory {
+    fn from(value: Vec<AclCategory>) -> Self {
+        AclCategory::Multi(value)
+    }
+}
+
+impl From<&str> for AclCategory {
+    fn from(value: &str) -> Self {
+        match value {
+            "" => AclCategory::None,
+            "keyspace" => AclCategory::Keyspace,
+            "read" => AclCategory::Read,
+            "write" => AclCategory::Write,
+            "set" => AclCategory::Set,
+            "sortedset" => AclCategory::SortedSet,
+            "list" => AclCategory::List,
+            "hash" => AclCategory::Hash,
+            "string" => AclCategory::String,
+            "bitmap" => AclCategory::Bitmap,
+            "hyperloglog" => AclCategory::HyperLogLog,
+            "geo" => AclCategory::Geo,
+            "stream" => AclCategory::Stream,
+            "pubsub" => AclCategory::PubSub,
+            "admin" => AclCategory::Admin,
+            "fast" => AclCategory::Fast,
+            "slow" => AclCategory::Slow,
+            "blocking" => AclCategory::Blocking,
+            "dangerous" => AclCategory::Dangerous,
+            "connection" => AclCategory::Connection,
+            "transaction" => AclCategory::Transaction,
+            "scripting" => AclCategory::Scripting,
+            _ if !value.contains(" ") => AclCategory::Single(value.to_string()),
+            _ => AclCategory::Multi(value.split_whitespace().map(AclCategory::from).collect()),
+        }
+    }
+}
+
+impl From<AclCategory> for String {
+    fn from(value: AclCategory) -> Self {
+        match value {
+            AclCategory::None => "".to_string(),
+            AclCategory::Keyspace => "keyspace".to_string(),
+            AclCategory::Read => "read".to_string(),
+            AclCategory::Write => "write".to_string(),
+            AclCategory::Set => "set".to_string(),
+            AclCategory::SortedSet => "sortedset".to_string(),
+            AclCategory::List => "list".to_string(),
+            AclCategory::Hash => "hash".to_string(),
+            AclCategory::String => "string".to_string(),
+            AclCategory::Bitmap => "bitmap".to_string(),
+            AclCategory::HyperLogLog => "hyperloglog".to_string(),
+            AclCategory::Geo => "geo".to_string(),
+            AclCategory::Stream => "stream".to_string(),
+            AclCategory::PubSub => "pubsub".to_string(),
+            AclCategory::Admin => "admin".to_string(),
+            AclCategory::Fast => "fast".to_string(),
+            AclCategory::Slow => "slow".to_string(),
+            AclCategory::Blocking => "blocking".to_string(),
+            AclCategory::Dangerous => "dangerous".to_string(),
+            AclCategory::Connection => "connection".to_string(),
+            AclCategory::Transaction => "transaction".to_string(),
+            AclCategory::Scripting => "scripting".to_string(),
+            AclCategory::Single(s) => s,
+            AclCategory::Multi(v) => v
+                .into_iter()
+                .map(String::from)
+                .collect::<Vec<_>>()
+                .join(" "),
+        }
+    }
+}
+
+impl std::fmt::Display for AclCategory {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", String::from(self.clone()))
+    }
+}
+
 /// The values allowed in the "info" sections and dictionaries.
 #[derive(Debug, Clone)]
 pub enum InfoContextBuilderFieldBottomLevelValue {
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -135,6 +135,9 @@ macro_rules! redis_module {
         data_types: [
             $($data_type:ident),* $(,)*
         ],
+        // eg: `acl_category: "name_of_module_acl_category",`
+        // This will add the module to the specified (optional) ACL category.
+        // All commands will inherit this category.
         $(acl_category: $module_acl_categories:expr,)?
         $(init: $init_func:ident,)* $(,)*
         $(deinit: $deinit_func:ident,)* $(,)*
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -7,7 +7,8 @@ macro_rules! redis_command {
      $firstkey:expr,
      $lastkey:expr,
      $keystep:expr,
-     $acl_categories:expr) => {{
+     $acl_categories:expr
+    ) => {{
         let name = CString::new($command_name).unwrap();
         let flags = CString::new($command_flags).unwrap();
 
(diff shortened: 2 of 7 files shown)


eranhd:
please elaborate what is this PR come to fix

src/macros.rs
```diff
@@ -270,18 +271,25 @@ macro_rules! redis_module {
...
-                        raw::redis_log(ctx, &format!("Error: failed to add ACL category {}", $acl_category));
+                    if RM_AddACLCategory(ctx, module_acl_categories.as_ptr()) == raw::Status::Err as c_int {
+                        raw::redis_log(ctx, &format!("Error: failed to add ACL category {}", $module_acl_categories));
                         return raw::Status::Err as c_int;
                     }
                 }
-            )*
+            )?
 
+            let module_acl_categories = module_acl_categories.to_str().unwrap();
```
eranhd, line 285: add a comment with an example of how module_acl_categories should looks like

ephraimfeldblum:
> please elaborate what is this PR come to fix

- it fixes the problem of naming the meta-variable class `$acl_category` identically to its identifier.
- it reduces code duplication on the module side (eg `RedisJSON`), all commands from the module are given the ACL category of the module without having to be expressly listed per command.
- it adds a test on the `redismodule-rs` side to ensure it works as expected.

eranhd (approved):

eranhd (approved):

MeirShpilraien:
Please modify ACL test to verify the new functionality, and also add this option to the command proc macro (and verify it with test)

src/macros.rs
```diff
@@ -7,7 +7,9 @@ macro_rules! redis_command {
      $firstkey:expr,
      $lastkey:expr,
      $keystep:expr,
-     $acl_categories:expr) => {{
+     $mandatory_acl_categories:expr
+     $(, $optional_acl_categories:expr)?
+    ) => {{
```
iddm, line 12: For the simplicity, I'd do two things:

1. Add a separate enum: `CommandAclCategory`:
```rust
enum CommandAclCategory {
    Write,
    Read,
    Fast,
    Other(String),
}

impl std::fmt::Display for CommandAclCategory {
    // ...
}
```

  Where we would add all the already existing, commonly used acl categories which have already existed for a long time, and would always specifying a new one by having the `Other` variant.

2. If the mandatory ACL category is mandatory, why not simply omit it and add it automatically? We are not in C, so we can have two macros, with it and without. In the "without" macro branch, I'd invoke the "with" branch with the mandatory ACL category name as the module name (this is the case for all the modules we are currently adding the ACLs anyway).
3. Rename the `acl_categories` arguments, as we specify them separately, not all within one argument: `mandatory_acl_category` and `optional_acl_category`.

src/macros.rs
```diff
@@ -37,34 +39,65 @@ macro_rules! redis_command {
...
-                return $crate::raw::Status::Err as c_int;
-            }
+        if let Some(RM_SetCommandACLCategories) = $crate::raw::RedisModule_SetCommandACLCategories {
+            let mut acl_categories = CString::new("").unwrap();
+            $(
+                if $mandatory_acl_categories != "" && $optional_acl_categories != "" {
+                    acl_categories = CString::new(format!("{} {}", $mandatory_acl_categories, $optional_acl_categories)).unwrap();
+                } else if $optional_acl_categories != "" {
+                    acl_categories = CString::new($optional_acl_categories).unwrap();
+                }
```
iddm, line 66: I don't like them being strings; I'd prefer having those as a type-safe `CommandAclCategory` enum, which would allow it to convert itself to a string, and a vector of those would be able to be joined with spaces and provide the same result. Just from the user perspective, the constant strings isn't the way to go, especially the number of times we repeat them.

src/macros.rs
```diff
@@ -37,34 +39,65 @@ macro_rules! redis_command {
...
+                // Fail if mandatory ACL categories are not set.
                 if RM_SetCommandACLCategories(command, acl_categories.as_ptr())
                     == $crate::raw::Status::Err as c_int
                 {
                     $crate::raw::redis_log(
                         $ctx,
                         &format!(
-                            "Error: failed to set command {} ACL categories {}",
-                            $command_name, $acl_categories
+                            "Error: failed to set command `{}` mandatory ACL categories `{}`",
```
iddm, line 88: I have forgotten; don't the macro arguments work well with the format arguments just yet, so can't they be inlined? I think the dollar sign breaks the parser there.

src/macros.rs
```diff
@@ -37,34 +39,65 @@ macro_rules! redis_command {
...
+            return $crate::raw::Status::Err as c_int;
+        }
 
-            let command =
-                unsafe { $crate::raw::RedisModule_GetCommand.unwrap()($ctx, name.as_ptr()) };
-            if command.is_null() {
-                return $crate::raw::Status::Err as c_int;
-            }
+        if let Some(RM_SetCommandACLCategories) = $crate::raw::RedisModule_SetCommandACLCategories {
+            let mut acl_categories = CString::new("").unwrap();
```
iddm, line 60: I wonder if we could simply use `CString::default` here with the same result?

src/macros.rs
```diff
@@ -146,7 +181,8 @@ macro_rules! redis_module {
                 $firstkey:expr,
                 $lastkey:expr,
                 $keystep:expr,
-                $acl_categories:expr
+                $mandatory_command_acl_categories:expr
```
iddm, line 184: Again, this is plural but represents just one category; IIRC, there is just one mandatory category now, or am I wrong? Either way, I understand the precaution here of actually passing more than one as a string separated with spaces, but I'd really like to have it more verbose and explicit by having the enum and a `[,]` expression instead then.

tests/integration.rs
```diff
@@ -329,6 +329,39 @@ fn test_get_current_user() -> Result<()> {
...
+}
+
+#[test]
+#[cfg(feature = "min-redis-compatibility-version-8-0")]
+fn test_set_acl_categories_commands() -> Result<()> {
+    let port: u16 = 6490;
+    let _guards = vec![start_redis_server_with_module("acl", port)
+        .with_context(|| "failed to start redis server")?];
+    let mut con =
+        get_redis_connection(port).with_context(|| "failed to connect to redis server")?;
```
iddm, line 354: Haven't we merged the changes of mine where for all of this we just have a single function instead? Somewhere in `tests/utils.rs` or something?

iddm: https://github.com/RedisLabsModules/redismodule-rs/pull/376/files here it is @MeirShpilraien please merge it, you approved it the last time and the freebsd failed, I guess, only due to their repositories having been shut down, or it got stuck, or something else which was usual for it to break.

src/macros.rs
ephraimfeldblum: the `$` does break the parser, they can't be inlined (as macro args)

src/macros.rs
ephraimfeldblum: the point is to allow the user to provide any arbitrary categories they like. we cannot provide an exhaustive enum, because the set of categories is infinite

src/macros.rs
ephraimfeldblum: 1. doable. will take under advisement.
2. there is no one singular mandatory ACL category. the mandatory ACL categories are those that are mandatory for the module to work at all, and loading the module should fail if they are not added. optional ACL categories are those that we warn on failure to add, but do not fail.
3. we do specify multiple categories within one argument. as a space delimited string.

src/macros.rs
ephraimfeldblum: there may be many mandatory categories. mandatory just means that if adding any of these categories fails, the module loading should fail.

tests/integration.rs
ephraimfeldblum: i do not see it there.

iddm: Yes, we need to merge the PR for these changes to arrive. False alarm for now, but if we merge it before this one, then it would be nice to use it here as well.

src/macros.rs
iddm: Alright, then  I'd like to have those not as string literals but as a comma-separated values list, which implements `From<T> for CommandAclCategory` enum either way, so either a string, or a `CommandAclCategory` enum variant itself.

src/macros.rs
iddm: :-( Thanks

src/macros.rs
iddm: Thanks for the response!

How we want to specify them for redis - this is the implementation detail. The interface here is in Rust, and we should allow the safest, most idiomatic way possible, which is an explicit set of type-safe values. Do you agree? For the user of this macro it shouldn't matter how exactly those end up being sent to the server, the user shouldn't care about this, nor should he know about this, if this is hidden behind the interface. This is not why this crate is simply forwarding the rust types to the C-api, right?

tests/integration.rs
iddm: We have just merged the PR: https://github.com/RedisLabsModules/redismodule-rs/pull/376

You can use the function there to simplify the test code for the new tests.

src/macros.rs
iddm: Yes, okay, that I am not discussing anymore. I am talking now about the type safety and the explicit lists.

src/context/mod.rs
```diff
@@ -926,6 +926,113 @@ bitflags! {
...
+impl From<Vec<AclCategory>> for AclCategory {
+    fn from(value: Vec<AclCategory>) -> Self {
+        AclCategory::Multi(value)
+    }
+}
+
+impl From<&str> for AclCategory {
+    fn from(value: &str) -> Self {
+        match value {
+            "" => AclCategory::None,
```
iddm, line 967: What should the result be in the case of ` read` or `read `? Is it even possible?

src/macros.rs
```diff
@@ -56,13 +58,15 @@ macro_rules! redis_command {
...
+            let mut acl_categories = CString::default();
             $(
-                if $mandatory_acl_categories != "" && $optional_acl_categories != "" {
-                    acl_categories = CString::new(format!("{} {}", $mandatory_acl_categories, $optional_acl_categories)).unwrap();
-                } else if $optional_acl_categories != "" {
-                    acl_categories = CString::new($optional_acl_categories).unwrap();
+                let optional_acl_categories = AclCategory::from($optional_acl_categories);
+                if mandatory_acl_categories != AclCategory::None && optional_acl_categories != AclCategory::None {
+                    acl_categories = CString::new(format!("{} {}", mandatory_acl_categories, optional_acl_categories)).unwrap();
+                } else if optional_acl_categories != AclCategory::None {
```
iddm, line 68: I'd personally add a method for this `AclCategory::is_none()` and `is_some()`. It may clash with the interface of `Option` and confuse the reader, though, but there is no better alternative.

src/macros.rs
```diff
@@ -56,13 +58,15 @@ macro_rules! redis_command {
...
+                    acl_categories = CString::new(format!("{}", $optional_acl_categories)).unwrap();
```
iddm, line 69: Can we just `$optional_acl_categories.to_string()`?

iddm (approved):
Nice! I don't strictly require any changes, just suggesting a few more things, but this is good to go.

src/macros.rs
ephraimfeldblum: it needs to be a `c_string` to pass to `Redis`

src/context/mod.rs
ephraimfeldblum: error?
passing that to Redis will be rejected

src/macros.rs
iddm: I know, I meant instead of using the `format!` macro we could have simply used the `to_string()` method of the `fmt::Display` trait, no?

PR #419, opened 2025-08-28
Encapsulates RedisModule_ScanKey as an Rust Iterator

DarthB:
WIP


src/context/key_cursor.rs
```diff
@@ -0,0 +1,280 @@
...
+    // we're not the owner of field and value strings
+    field.take();
+    value.take();
+
+    key.take(); // we're not the owner of the key either
+}
+
+/// The callback that is used inside the iterator variant accessible via [`ScanKeyCursor::iter`] for `RedisModule_ScanKey`. 
+/// 
+/// It buffers copies of the field and value strings as the lifetime of them ends when with the call
```
JonasKruckenberg, line 237: do we know this for sure?

src/context/key_cursor.rs
```diff
@@ -0,0 +1,280 @@
...
+) {
+    // `data` is a stack slot
+    let slot = data.cast::<StackSlot>();
+    let slot = &mut (*slot);
+
+    // todo: use new-type with refcount handling for better performance, otherwise we have to copy at this point
+    // we know that this callback will be called in a loop scope and that we 
+    // need to store the RedisString longer than the ScanKey invocation but not much
+    // longer and in case of batched results we don't need to store everything in memory
+    // but only the last batch.
```
JonasKruckenberg, line 255: can't leave todos in code

src/context/key_cursor.rs
```diff
@@ -0,0 +1,280 @@
...
+    let field = raw::RedisModule_CreateStringFromString.unwrap()(slot.ctx.get_raw(), field);
+    let value = raw::RedisModule_CreateStringFromString.unwrap()(slot.ctx.get_raw(), value);
+
+    let field = RedisString::from_redis_module_string(slot.ctx.get_raw(), field);
```
JonasKruckenberg, line 259: This will not actually do what you want it to (prevent the strings from being freed) you need to call [`RedisModule_RetainString`](https://github.com/redis/redis/blob/8ad5421502241d6088b701bb4a4262124343345a/src/module.c#L2776-L2810) on the `RedisModuleString` for that ([`safe_clone`](https://docs.rs/redis-module/latest/redis_module/struct.RedisString.html#method.safe_clone) in Rust).

JonasKruckenberg (changes requested):
Generally on the right track, just a couple of things:

1. The API should be *either* callback-based or iterator-based, never both. This will just lead to confusion as to which API is the correct one. After we resolve the soundness issues below we should be able to just have the iterator too anyways.
2. The Iterator needs to extend the lifetime of the strings beyond the callback. We should just do that with the `RetainString`, buffering doesn't help much.
3. `RetainString` is not thread-safe. We must ensure the GIL is held while calling it. Can we know if the GIL is held? Do we need to require callers to ensure this? EDIT: yes we know the GIL is held, since we are passed the `*const RedisModuleCtx`.
4. Since the Cursor is `Send` strings can be sent to other threads. According to the  `RetainString` doc comment we must call `RedisModule_TrimStringAllocation` to obtain a thread-safe string pointer.
5. According to the [`RedisModule_ScanKey`](https://github.com/redis/redis/blob/8ad5421502241d6088b701bb4a4262124343345a/src/module.c#L11339-L11387) docs, the `value` ptr can actually be null, we need to account for that.
6. ~~EDIT: The first arguments to the `RedisModule_Scan` callback is actually a `*const RedisModuleCtx` not a `*mut RedisModuleKey` where does this come from exactly? We need to change this wherever it came from.~~
7. ~~EDIT: reading through the redis source a bit more, I think it is safe to assume that `RedisModule_Scan` will never call the callback more than once for a given invocation. We don't need buffering of returned values for this either.~~

Once we have all this, I would suggest stripping back the implementation to what we had in https://github.com/RediSearch/RediSearch/pull/6509 (as simple as possible) and only exposing an Iterator interface (implemented directly on the cursor?).

src/context/key_cursor.rs
```diff
@@ -0,0 +1,129 @@
...
+
+        let callback = unsafe { &mut *(data.cast::<F>()) };
+        callback(&key, &field, &value);
+
+        // we're not the owner of field and value strings
+        field.take();
+        value.take();
+
+        key.take(); // we're not the owner of the key either
+    }
```
JonasKruckenberg, line 128: just move this into the body of the `scan` method, so its private by default and we colocate the tricky implementation details

src/context/key_cursor.rs
```diff
@@ -0,0 +1,122 @@
...
+        let inner_cursor = unsafe { raw::RedisModule_ScanCursorCreate.unwrap()() };
+        Self { key, inner_cursor }
+    }
+
+    /// Restarts the cursor from the beginning.
+    pub fn restart(&self) {
+        unsafe { raw::RedisModule_ScanCursorRestart.unwrap()(self.inner_cursor) };
+    }
+
+    pub fn scan<F: FnMut(&RedisKey, &RedisString, &RedisString)>(&self, f: F) -> bool {
```
JonasKruckenberg, line 66: docs

src/context/key_cursor.rs
```diff
@@ -0,0 +1,122 @@
...
+            field.take();
+            value.take();
+
+            key.take(); // we're not the owner of the key either
+        }
+
+        // Safety: The c-side initialized the function ptr and it is is never changed,
+        // i.e. after module initialization the function pointers stay valid till the end of the program.
+        let res = unsafe {
+            raw::RedisModule_ScanKey.unwrap()(
```
JonasKruckenberg, line 99:
```suggestion
// lift this out of the `scan_method`
let scan_key = raw::RedisModule_ScanKey.unwrap();
            
            scan_key(
                self.key.key_inner,
                self.inner_cursor,
                Some(scan_callback::<F>),
                &f as *const F as *mut c_void,
            )
```

You should lift the potentially panicking branch here out of the `scan` method

src/context/key_cursor.rs
```diff
@@ -0,0 +1,122 @@
...
+        Self { key, inner_cursor }
+    }
+
+    /// Restarts the cursor from the beginning.
+    pub fn restart(&self) {
+        unsafe { raw::RedisModule_ScanCursorRestart.unwrap()(self.inner_cursor) };
+    }
+
+    pub fn scan<F: FnMut(&RedisKey, &RedisString, &RedisString)>(&self, f: F) -> bool {
+        // The following is the callback definition. The callback may be called multiple times per `RedisModule_ScanKey` invocation.
```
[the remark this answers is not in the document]
DarthB: https://github.com/RedisLabsModules/redismodule-rs/pull/419/commits/c59dbaa04569bc3779953ffc4aef4d6f5cea0a71

src/context/key_cursor.rs
```diff
@@ -0,0 +1,122 @@
...
+        // The callback is used by [`ScanKeyCursor::scan`] and [`ScanKeyCursor::for_each`] as argument to `RedisModule_ScanKey`.
+        //
+        // The `data` pointer is the closure given to [`ScanKeyCursor::scan`] or [`ScanKeyCursor::for_each`]. 
+        // The callback forwards references to the key, field and value to that closure.
```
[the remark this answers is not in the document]
DarthB: https://github.com/RedisLabsModules/redismodule-rs/pull/419/commits/c59dbaa04569bc3779953ffc4aef4d6f5cea0a71

PR #428, opened 2025-09-24
Revamp the CI pipeline, including publishing

LukeMathWalker:
For releasing, we switched to [`release-plz`](). It handles version updates in manifests, changelog maintainance and crates.io publishing, thus simplifying the overall workflow.

I'd like to avoid using an API token for publishing. We can instead rely on [trusted publishing from GitHub Actions](https://crates.io/docs/trusted-publishing). The only prerequisite is a bit of configuration on the crates.io management panel (cc @oshadmi).


Cargo.toml
```diff
@@ -1,11 +1,19 @@
+[workspace]
+members = [".", "redismodule-rs-macros", "redismodule-rs-macros-internals"]
```
enricozb, line 2: is it normal to have this root member like this?

LukeMathWalker: I generally prefer to have all root members in subfolders, since it makes `cargo build` in the root of the repository equivalent to `cargo build --workspace`. 
But I know that some projects prefer to have the "key" package at the root level, so no strong preferences here.

.github/workflows/release.yml
```diff
@@ -0,0 +1,34 @@
+name: Release-plz
+
+on:
+  push:
+    branches:
+      - master
```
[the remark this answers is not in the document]
oshadmi: @LukeMathWalker Will this support multiple concurrent releases? ☝🏼

oshadmi:
@LukeMathWalker using #429 #430 I could add myself as owner of redis-module and we could try setting up Trusted Publishing

.github/workflows/release.yml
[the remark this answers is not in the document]
LukeMathWalker: When looking at the [version history](https://crates.io/crates/redis-module/versions), this didn't seem like a requirement—from what I can see, no backports have been made in the last few years.

I would recommend the following setup to minimise complexity:
- Use `master` as trunk and cut releases for the latest major from there (i.e. all `2.x.y`, until `3.0.0` is released)
- Branch out when a new major release goes out (i.e. create a `v2` branch once `master` moves over to `3.z.w`)

If we need to backport something, we merge PRs into the backport branch and use something like [`cargo-release`](https://github.com/crate-ci/cargo-release) with a short-lived token to cut a release. In the meantime, I can look into adding support for branches to `release-plz`.

test.sh
```diff
@@ -1,2 +1,2 @@
 #!/usr/bin/env sh
-cargo test --all --all-targets --no-default-features --features min-redis-compatibility-version-7-4
+cargo test --all --all-targets --no-default-features --features min-redis-compatibility-version-8-0
```
oshadmi, line 2: @LukeMathWalker Why are we increasing the min compatibility?

LukeMathWalker: There were examples referring to `min-redis-compatibility-version-8-0`, but the feature flag didn't actually exist. I added it and updated test.sh to use it, since it seemed to be designed to test the latest version. 
CI still tests all versions (`test.sh` is never invoked anywhere from what I can see, it's just a local dev helper).

.github/workflows/ci.yml
```diff
@@ -33,35 +19,34 @@ jobs:
           - [7-2, "7.2"]
```
oshadmi, line 19: @LukeMathWalker Should we add 8-0 here? Or unrelated to this PR?

LukeMathWalker: We could add it now, but since I don't see 7.4 there either, I'm inclined to defer to another PR

PR #435, opened 2025-12-10
fix: add `to_raw_parts` method & fix unsoundness bug

JonasKruckenberg:
This change introduces the `RedisKey::to_raw_parts` method as a counterpart to `RedisKey::from_raw_parts`. We need to be able to access the inner pointer for more low-level operations. It also marks `from_raw_parts` as unsafe adding documentation to both.

Additionally this fixes an unsoundness bug in in `KeyCursor` that was leading to possible use-after free of the key name provided to the scan callback.

### DIFF
--- a/examples/scan_keys.rs
+++ b/examples/scan_keys.rs
@@ -17,8 +17,8 @@ fn scan_keys(ctx: &Context, _args: Vec<RedisString>) -> RedisResult {
     let cursor = KeysCursor::new();
     let mut res = Vec::new();
 
-    let scan_callback = |_ctx: &Context, key_name: RedisString, _key: Option<&RedisKey>| {
-        res.push(RedisValue::BulkRedisString(key_name));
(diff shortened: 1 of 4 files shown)


GlenDC (approved):
Looks good.

Only thing that concerns me, but outside scope of this PR I suppose, is the lack of sanity checks to at least have some guarnatees for these safety checks.

I also wonder if it should not be better to have a borrow type specific for these purposes which doesn't do any dropping at all so that you cannot even forget to disable the destructors for cases like this...

JonasKruckenberg:
> Only thing that concerns me, but outside scope of this PR I suppose, is the lack of sanity checks to at least have some guarnatees for these safety checks.
> 
> I also wonder if it should not be better to have a borrow type specific for these purposes which doesn't do any dropping at all so that you cannot even forget to disable the destructors for cases like this...

agreed, but squarely outside the scope of this PR

PR #436, opened 2025-12-10
feat: add `RedisKey::name_raw` method

JonasKruckenberg:
This change adds the `RedisKey::name_raw` method which allows access to the raw `RedisModuleString` name of this key.

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -87,6 +87,28 @@ impl RedisKey {
         Self { ctx, key_inner }
     }
 
+    /// Returns a raw pointer to the name of this key.
+    ///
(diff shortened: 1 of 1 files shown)


src/key.rs
```diff
@@ -120,6 +109,24 @@ impl RedisKey {
         self.key_inner == null_key
     }
 
+    /// # Panics
+    ///
+    /// Will panic if `RedisModule_ModuleTypeGetValue` is missing in redismodule.h
```
GlenDC: Why is that a panic and not a compile error?

GlenDC (approved):
LGTM, just minor question.

PR #461, opened 2026-04-22
[MOD-13433] Add RedisModuleIO wrapper for type-safe RDB serialization/deserialization

chesedo:
This makes a new `io::RedisModuleIO` to replace the need for using the raw pointer to `raw::RedisModuleIO` for RDB operations

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -1,5 +1,5 @@
+use crate::error::Error;
 use crate::raw;
-use crate::rediserror::RedisError;
 use crate::RedisBuffer;
 
 /// Wrapper around RedisModuleIO for type-safe serialization/deserialization.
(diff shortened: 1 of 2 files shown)


JoanFM (dismissed):

src/io.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub fn write_string(&mut self, s: &str) {
+        raw::save_string(self.io, s);
+    }
+
+    /// Reads an unsigned integer from the IO stream.
+    ///
+    /// # Errors
+    /// Returns an error if the operation fails or if an IO error occurred.
+    pub fn read_unsigned(&mut self) -> Result<u64, RedisError> {
+        let val = raw::load_unsigned(self.io)?;
```
GuyAv46: We not pass the returned `Result` directly?
```
    pub fn read_unsigned(&mut self) -> Result<u64, RedisError> {
        raw::load_unsigned(self.io)
    }
```

chesedo: We should be able to. Let me check...

chesedo: Done in https://github.com/RedisLabsModules/redismodule-rs/pull/461/commits/411aea04da6deb400b52b94afe416f2e245c25bc

GuyAv46 (approved):

PR #466, opened 2026-05-11
ci: Add a dry-run workflow for release PRs

LukeMathWalker:


.github/workflows/release-pr-dry-run.yml
```diff
@@ -0,0 +1,30 @@
...
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  # Verify that all workspace crates are publishable, so that the actual
+  # release (triggered by merging the release PR) doesn't fail mid-way.
+  publish-dry-run:
+    name: Dry-run `cargo publish` for the workspace
+    if: startsWith(github.head_ref, 'release-plz-')
```
oshadmi: @LukeMathWalker This only runs for release-plz PRs, so it catches publishability issues at release-PR time rather than when a normal PR introduces them.
If we want this as a regression check, it should probably run on all PRs to master (and also version branches in addition to master?)
Maybe add a path-filter to Cargo.toml/Cargo.lock/src changes

LukeMathWalker: We could, but it's honestly rare to break a working publishing set-up. That's why I normally reserve this dry-run as a pre-publish sanity check.
But if you prefer we can make it part of the general CI flow.

oshadmi: I prefer to avoid surprises found only when we need to release/publish

LukeMathWalker: Changed!

Issue #472, opened 2026-07-10
redis_event_handler! generated callback panics on non-UTF-8 event names


joshrotenberg:
## Summary

The callback generated by `redis_event_handler!` converts the event name with `to_str().unwrap()`. A keyspace event whose name is not valid UTF-8 panics inside the generated handler, before the module author's handler runs.

## Location

`src/macros.rs` line 132 (master, 7fb620f):

```rust
let event_str = unsafe { CStr::from_ptr(event) };
$event_handler(
    &context,
    $crate::NotifyEvent::from_bits_truncate(event_type),
    event_str.to_str().unwrap(),
    redis_key,
);
```

## Reproduction path

Any co-loaded module can fire an arbitrary event name via `RedisModule_NotifyKeyspaceEvent` (the API takes a C string, so anything NUL-free is possible, including invalid UTF-8 byte sequences). A module built on this crate that subscribes with `event_handlers:` or `redis_event_handler!` then panics inside the generated `extern "C"` callback. A panic across an FFI boundary into Redis is undefined behavior in older Rust editions and at best aborts the server process.

Built-in event names are ASCII, so this does not trigger in normal operation; the exposure is a hostile or buggy co-loaded module taking down the server through any subscriber built on this crate.

## Suggested fix

Either of:

- Lossy decode: `String::from_utf8_lossy(event_str.to_bytes())`, passing `&str` as today with replacement characters for invalid sequences. Non-breaking for handler signatures.
- Bytes-based handler signature (`&[u8]` for the event name, matching how the key is already passed as `&[u8]`). Breaking, but consistent with the key parameter.

The lossy decode is a two-line change and preserves the public API.

oshadmi:
I would take the lossy option and not change the API. The event name is string-like, not raw/blob data: Redis passes it as `const char *` with no length, so it must be NUL-terminated and is not binary-safe like `RedisModuleString`. Lossy decoding avoids panicking in the event callback while preserving the existing `&str` API.

joshrotenberg:
Fixed by #475, merged to master in 591e7db.

`redis_event_handler!` and `redis_command!` now decode with `to_string_lossy()` instead of `to_str().unwrap()`, so non-UTF-8 event names and ACL categories no longer panic across the FFI boundary. Covered by `test_key_space_notification_with_invalid_utf8_event_name` in tests/integration.rs.

[...]

PR #475, opened 2026-07-14
Allow non-UTF-8 strings in event names and ACL categories

memark:
Two places in macros.rs called `CStr::to_str().unwrap()` which would panic for invalid UTF-8. Changed both places to call `CStr::to_string_lossy()` instead.

NB: For invalid UTF-8 this call will allocate memory.

Solves #472.

### DIFF
--- a/examples/events.rs
+++ b/examples/events.rs
@@ -1,11 +1,15 @@
 use redis_module::{
-    redis_module, Context, NotifyEvent, RedisError, RedisResult, RedisString, RedisValue, Status,
+    raw, redis_module, Context, NotifyEvent, RedisError, RedisResult, RedisString, RedisValue,
+    Status,
 };
+use std::ffi::CString;
 use std::ptr::NonNull;
 use std::sync::atomic::{AtomicI64, Ordering};
+use std::sync::Mutex;
 
 static NUM_KEY_MISSES: AtomicI64 = AtomicI64::new(0);
 static NUM_KEYS: AtomicI64 = AtomicI64::new(0);
+static LAST_GENERIC_EVENT: Mutex<String> = Mutex::new(String::new());
 
 fn on_event(ctx: &Context, event_type: NotifyEvent, event: &str, key: &[u8]) {
     if key == b"num_sets" {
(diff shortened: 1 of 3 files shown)


src/macros.rs
```diff
@@ -77,7 +77,7 @@ macro_rules! redis_command {
                         $ctx,
                         &format!(
                             "Warning: failed to set command `{}` ACL categories `{}`",
-                            $command_name, acl_categories.to_str().unwrap()
+                            $command_name, &acl_categories.to_string_lossy()
```
oshadmi, line 80: Should ACL categories also be mentioned in the PR title? or a more generic title, e.g., Avoid UTF-8 panics in macro C-string handling

src/macros.rs
oshadmi, line 1: Can add a test with invalid UTF-8?

src/macros.rs
memark: Good catch, updated PR title.

src/macros.rs
memark: Added a test for an invalid UTF-8 event name.

The ACL category currently cannot be invalid UTF-8 (but I still think the change is reasonable), so no test for that.

examples/events.rs
```diff
@@ -45,6 +49,39 @@ fn event_send(ctx: &Context, args: Vec<RedisString>) -> RedisResult {
...
+    if args.len() > 1 {
+        return Err(RedisError::WrongArity);
+    }
+
+    let key_name = RedisString::create(NonNull::new(ctx.ctx), "mykey");
+    // Fire a keyspace event whose name is not valid UTF-8 (0xFF can never
+    // appear in a UTF-8 string), the way a C module could. This has to go
+    // through the raw API because the safe wrapper only accepts &str.
+    let event = CString::new(&b"ev\xFFnt"[..]).unwrap();
+    let status: Status = unsafe {
```
oshadmi, line 72: Nit: Is a ``// SAFETY:`` comment needed? (might be copied over to code?)

oshadmi (dismissed):

examples/events.rs
memark: Sure thing, added one!

oshadmi (approved):

PR #483, opened 2026-08-17
[MOD-17680] Honour layouts requiring more than 16-byte alignment

memark:
Tracked as [MOD-17680](https://redislabs.atlassian.net/browse/MOD-17680).

## Problem

`RedisAlloc::alloc` rounds the requested **size** up to the layout's alignment, but never
obtains an aligned **address**:

```rust
let size = (layout.size() + layout.align() - 1) & (!(layout.align() - 1));
alloc(size).cast()
```

`RedisModule_Alloc` is `zmalloc`, which returns `max_align_t`-aligned memory — 16 bytes on
the platforms this crate targets. Any `Layout` asking for more than that is silently
served under-aligned, which is undefined behaviour. In a build with debug assertions the
first dereference of such a pointer trips
`core::panicking::panic_misaligned_pointer_dereference` and the process aborts; in a
release build it is a latent SIGSEGV on instructions that require alignment.

Whether it is observed depends entirely on what the platform allocator happens to return:

- **Linux** — jemalloc's size classes usually return a sufficiently aligned address by
  accident, so the defect stays hidden.
- **aarch64** — over-aligned types are rarer, so the code path is often never taken.
- **macOS x86_64** — neither applies, and it aborts.

We hit this in RediSearch. `memchr::memmem::Finder` embeds a `packedpair::Finder<__m256i>`
on AVX2 targets, giving it `align = 32`; boxing one and dereferencing it aborts the server
on every `CONTAINS` query. Dispatching the same test suite twice on a macOS x86_64 runner
gave 22 misaligned-pointer aborts and 19 failed tests without a fix, versus 0 and 0 with
one, over the same 4355 executed tests. That run validated an equivalent fix rather than
this diff: a downstream crate wrapping `RedisAlloc` with the same
over-allocate-and-store-the-header scheme. This PR folds the scheme into `RedisAlloc`
itself, so the algorithm is the one that was measured but the code is not.

## Change

The module API has no aligned-allocation primitive — `RedisModule_Alloc`, `Calloc`,
`Realloc`, `TryAlloc` and `MallocSize` all take or report a plain size — so an
over-aligned block has to be carved out of a larger one by hand. `alloc` now
over-allocates by `align + size_of::<*mut u8>()`, returns the first suitably aligned
address at or above the header, and stores the pointer that must be given back to
`RedisModule_Free` in the word directly below it. `dealloc` reads that word back.

Two things are deliberate:

- **Layouts within `MIN_ALIGN` keep the existing path, size rounding included.** That
  rounding is load-bearing: for `Layout(size = 8, align = 16)` a bare `malloc(8)` under
  jemalloc may return only 8-aligned memory, so removing it would introduce on Linux
  exactly the bug this PR fixes on macOS. There is a regression test for it.
- **The header stores the pointer, not the offset**, so the provenance Redis handed out is
  preserved and `dealloc` returns the very pointer `alloc` received.

`dealloc` has to start reading its `layout` argument (it was `_layout`) to know which
branch `alloc` took, so the two cannot be changed independently.

`realloc` and `alloc_zeroed` are not overridden, and the `GlobalAlloc` defaults compose
correctly with this: they are defined in terms of `alloc`/`dealloc`, so an over-aligned
block is reallocated as allocate-copy-free rather than through `RedisModule_Realloc`.
That is a small cost paid only by over-aligned layouts, which are rare.

## Compatibility

One behavioural note worth flagging: an over-aligned block now has a header below it, so a
pointer allocated by Rust and freed directly by C with `RedisModule_Free` would free the
user address rather than the base. Layouts up to 16 bytes of alignment — every such
pointer that exists today, since the crate could not previously produce anything else
usable — are completely untouched by this patch and keep working as before.

`DefragContext` has the same seam: `defrag_alloc` goes through the global allocator, while
`defrag_realloc` passes its pointer to `RedisModule_DefragAlloc`, which expects one that
came straight from `RM_Alloc`. For an over-aligned layout those two no longer compose — the
header would have Redis defragment an interior pointer, and a block that moved would carry
a stale base into its new location. Nothing is affected today: the only user of that
pairing I could find is RedisJSON's `DefragCtxAllocator`, and every layout `ijson` sends
through it is an `align(8)` header extended with primitive arrays. Nor does anything that
worked before stop working — an over-aligned `defrag_alloc` used to return memory that was
under-aligned for its own layout, so it could not be dereferenced at all. Letting the
defrag path carry over-aligned blocks needs a way to tell global-allocator pointers from
raw `RM_Alloc` ones, which is a separate change.

## Testing

`alloc`/`dealloc` are thin wrappers over `alloc_aligned`/`dealloc_aligned`, which take the
underlying allocator as a closure. That is what makes them testable without a running
Redis server behind the `RedisModule_*` function pointers; the tests substitute a
`zmalloc` double that mimics its contract — takes only a size, always returns
`MIN_ALIGN`-aligned memory — and asserts it is handed its own pointer back on free.

Five unit tests cover the fast path, the size-rounding regression, alignment and
usability of the over-aligned path across `[32, 64, 4096]`, freeing at the underlying
pointer, and null propagation. They pass under `cargo test` and under `cargo miri test`,
which is what actually checks the pointer arithmetic for provenance and out-of-bounds
access.

`cargo fmt --check` and `cargo clippy --workspace --all-targets` are clean (the one
remaining clippy warning is pre-existing, in `examples/proc_macro_commands.rs`).

### DIFF
--- a/src/alloc.rs
+++ b/src/alloc.rs
@@ -1,4 +1,5 @@
 use std::alloc::{GlobalAlloc, Layout};
+use std::ptr;
 
 use crate::raw;
 
@@ -18,6 +19,21 @@ fn allocation_free_panic(message: &'static str) -> ! {
 
 const REDIS_ALLOCATOR_NOT_AVAILABLE_MESSAGE: &str =
     "Critical error: the Redis Allocator isn't available.\n";
+
+/// The alignment `RedisModule_Alloc` provides on its own.
+///
+/// It is Redis' `zmalloc`, which forwards to an allocator handing out
+/// `max_align_t`-aligned addresses — jemalloc, tcmalloc or libc `malloc`
+/// depending on how the server was built.
+///
+/// Understating this costs an unnecessary fixup; overstating it hands out
+/// under-aligned memory, so it is the weakest guarantee those allocators share
+/// rather than the strongest any of them happens to provide.
+const MIN_ALIGN: usize = 2 * std::mem::size_of::<usize>();
+
+/// Size of the header kept directly below an over-aligned block, holding the
+/// pointer that has to be given back to `RedisModule_Free`.
+const HEADER: usize = std::mem::size_of::<*mut u8>();
 
 /// Defines the Redis allocator. This allocator delegates the allocation
 /// and deallocation tasks to the Redis server when available, otherwise
@@ -44,10 +60,15 @@ unsafe impl GlobalAlloc for RedisAlloc {
 }
 
 /// Allocate `layout` through `alloc`, which is expected to behave like
-/// `RedisModule_Alloc`: it takes a size, and returns a block aligned for any
-/// fundamental type and no more.
+/// `RedisModule_Alloc`: it takes a size, and returns a block aligned to at most
+/// [`MIN_ALIGN`].
 ///
-/// Returns a null pointer if `alloc` does.
+/// Layouts within [`MIN_ALIGN`] are served by `alloc` directly. A stricter one
+/// cannot be — the module API has no aligned-allocation primitive — so it is
+/// carved out of a larger block instead, with the underlying pointer stored in
+/// the [`HEADER`] word below the address returned to the caller.
+///
+/// Returns a null pointer if `alloc` does, or if the padded size overflows.
 ///
 /// Taking the allocator as a closure keeps this testable without a running
 /// Redis server behind `RedisModule_Alloc`.
@@ -56,20 +77,41 @@ unsafe impl GlobalAlloc for RedisAlloc {
 ///
 /// `alloc` must behave like [`GlobalAlloc::alloc`] for the size it is given.
 unsafe fn alloc_aligned(layout: Layout, alloc: impl FnOnce(usize) -> *mut u8) -> *mut u8 {
-    /*
-     * To make sure the memory allocation by Redis is aligned to the according to the layout,
-     * we need to align the size of the allocation to the layout.
-     *
-     * "Memory is conceptually broken into equal-sized chunks,
-     * where the chunk size is a power of two that is greater than the page size.
-     * Chunks are always aligned to multiples of the chunk size.
-     * This alignment makes it possible to find metadata for user objects very quickly."
-     *
-     * From: https://linux.die.net/man/3/jemalloc
-     */
-    let size = (layout.size() + layout.align() - 1) & (!(layout.align() - 1));
+    if layout.align() <= MIN_ALIGN {
+        /*
+         * To make sure the memory allocation by Redis is aligned to the according to the layout,
+         * we need to align the size of the allocation to the layout.
+         *
+         * "Memory is conceptually broken into equal-sized chunks,
+         * where the chunk size is a power of two that is greater than the page size.
+         * Chunks are always aligned to multiples of the chunk size.
+         * This alignment makes it possible to find metadata for user objects very quickly."
+         *
+         * From: https://linux.die.net/man/3/jemalloc
+         */
+        let size = (layout.size() + layout.align() - 1) & (!(layout.align() - 1));
 
-    alloc(size)
+        return alloc(size);
+    }
+
+    // Worst case the block starts one byte past an aligned address, so reaching
+    // the next one from just above the header costs `align - 1` further bytes.
+    let Some(size) = layout.size().checked_add(layout.align() + HEADER) else {
+        return ptr::null_mut();
+    };
+
+    let base = alloc(size);
+    if base.is_null() {
+        return base;
+    }
+
+    let offset = (base.addr() + HEADER).next_multiple_of(layout.align()) - base.addr();
+    let ptr = base.add(offset);
+    // Storing the pointer rather than the offset keeps the provenance Redis gave
+    // us, so `dealloc_aligned` hands back exactly what `alloc` returned.
+    ptr.cast::<*mut u8>().sub(1).write(base);
+
+    ptr
 }
 
 /// Free a pointer produced by [`alloc_aligned`] through the matching `free`.
@@ -78,7 +120,15 @@ unsafe fn alloc_aligned(layout: Layout, alloc: impl FnOnce(usize) -> *mut u8) ->
 ///
 /// `ptr` must have come from [`alloc_aligned`] with this same `layout`, and
 /// `free` must release blocks allocated by the closure that call was given.
-unsafe fn dealloc_aligned(ptr: *mut u8, _layout: Layout, free: impl FnOnce(*mut u8)) {
+unsafe fn dealloc_aligned(ptr: *mut u8, layout: Layout, free: impl FnOnce(*mut u8)) {
+    let ptr = if layout.align() <= MIN_ALIGN {
+        ptr
+    } else {
+        // `alloc_aligned` took the same branch for this layout, so the word below
+        // `ptr` is inside the allocation and holds the underlying pointer.
+        ptr.cast::<*mut u8>().sub(1).read()
+    };
+
     free(ptr);
 }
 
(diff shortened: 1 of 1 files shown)


src/alloc.rs
```diff
@@ -40,16 +91,259 @@ unsafe impl GlobalAlloc for RedisAlloc {
...
+    /// running server. Like `zmalloc` it takes only a size, and it returns the
+    /// weakest thing `zmalloc` may return: a block aligned to exactly
+    /// [`MIN_ALIGN`] and never more.
+    ///
+    /// That last part is deliberate. Real allocators over-align by chance often
+    /// enough that under-aligned allocations go unnoticed on Linux for years, and
+    /// a double inheriting that luck would pass these tests against an
+    /// implementation that never aligns anything.
+    #[derive(Default)]
+    struct Zmalloc {
```
daxhuiberts, line 165: I'd call it something like `TestAllocator` or something.

src/alloc.rs
```diff
@@ -40,16 +91,259 @@ unsafe impl GlobalAlloc for RedisAlloc {
...
+    let Some(size) = layout.size().checked_add(layout.align() + HEADER) else {
+        return ptr::null_mut();
+    };
+
+    let base = alloc(size);
+    if base.is_null() {
+        return base;
+    }
+
+    let offset = (base.addr() + HEADER).next_multiple_of(layout.align()) - base.addr();
```
daxhuiberts: This is the important line and looks correct!

src/alloc.rs
```diff
@@ -40,16 +91,259 @@ unsafe impl GlobalAlloc for RedisAlloc {
...
 
-        match raw::RedisModule_Alloc {
-            Some(alloc) => alloc(size).cast(),
-            None => allocation_free_panic(REDIS_ALLOCATOR_NOT_AVAILABLE_MESSAGE),
+        return alloc(size);
+    }
+
+    // Worst case the block starts one byte past an aligned address, so reaching
+    // the next one from just above the header costs `align - 1` further bytes.
+    let Some(size) = layout.size().checked_add(layout.align() + HEADER) else {
```
daxhuiberts, line 99: I think this alignment fix solution should also work with `layout.size().checked_add(layout.align())`. Knowing that `layout.align() => HEADER` we always have enough room to store the `HEADER` within that. This ensures allocations are never > 2*size.

I even expect the tests to still turn green when you change this line and not the tests.

src/alloc.rs
```diff
@@ -19,6 +20,21 @@ fn allocation_free_panic(message: &'static str) -> ! {
 const REDIS_ALLOCATOR_NOT_AVAILABLE_MESSAGE: &str =
     "Critical error: the Redis Allocator isn't available.\n";
 
+/// The alignment `RedisModule_Alloc` provides on its own.
+///
+/// It is Redis' `zmalloc`, which forwards to an allocator handing out
+/// `max_align_t`-aligned addresses — jemalloc, tcmalloc or libc `malloc`
+/// depending on how the server was built.
```
daxhuiberts, line 27: Redis internally in its code base calls it `zmalloc`. Let's start it with "Redis' allocator forwards ..." to not mix their internals in here.

src/alloc.rs
```diff
@@ -27,6 +43,41 @@ pub struct RedisAlloc;
...
+}
+
+/// Allocate `layout` through `alloc`, which is expected to behave like
+/// `RedisModule_Alloc`: it takes a size, and returns a block aligned to at most
+/// [`MIN_ALIGN`].
+///
+/// Layouts within [`MIN_ALIGN`] are served by `alloc` directly. A stricter one
+/// cannot be — the module API has no aligned-allocation primitive — so it is
+/// carved out of a larger block instead, with the underlying pointer stored in
+/// the [`HEADER`] word below the address returned to the caller.
```
daxhuiberts, line 69: It's a bit confusing to read, hopefully this one is better:

Layouts within [`MIN_ALIGN`] are served by `alloc` directly. Layouts with larger alignments are 'carved' out of a larger block instead to guarantee this larger alignment. The returned `alloc` pointer is stored in the [`HEADER`] word below the fully aligned address, and this fully aligned address is then returned to the caller.

src/alloc.rs
```diff
@@ -40,16 +91,259 @@ unsafe impl GlobalAlloc for RedisAlloc {
...
+    #[test]
+    fn a_failed_allocation_propagates_null() {
+        let layout = Layout::from_size_align(64, 64).unwrap();
+
+        let ptr = unsafe { alloc_aligned(layout, |_| ptr::null_mut()) };
+
+        // Null must come straight back out: writing a header below it would
+        // dereference the null pointer.
+        assert!(ptr.is_null());
     }
```
daxhuiberts, line 348: minor, but lets move this test up a bit with the rest. The previous test with the thread local and overriding `RedisModule_Alloc`/`RedisModule_Free` should be the last here.

src/alloc.rs
```diff
@@ -40,16 +91,259 @@ unsafe impl GlobalAlloc for RedisAlloc {
...
+        fn alloc(&self, requested: usize) -> *mut u8 {
+            // Room to push the block off a stricter boundary when it lands on one.
+            let layout = Layout::from_size_align(requested + MIN_ALIGN, MIN_ALIGN).unwrap();
+            let base = unsafe { System.alloc(layout) };
+
+            let offset = if base.addr() % (2 * MIN_ALIGN) == 0 {
+                MIN_ALIGN
+            } else {
+                0
+            };
```
daxhuiberts, line 179: It's not really clear what is happening here and why? I had Claude explain it to me and now I understand, but the code itself isn't explainatory enough.

src/alloc.rs
```diff
@@ -19,6 +20,21 @@ fn allocation_free_panic(message: &'static str) -> ! {
...
+/// The alignment `RedisModule_Alloc` provides on its own.
+///
+/// It is Redis' `zmalloc`, which forwards to an allocator handing out
+/// `max_align_t`-aligned addresses — jemalloc, tcmalloc or libc `malloc`
+/// depending on how the server was built.
+///
+/// Understating this costs an unnecessary fixup; overstating it hands out
+/// under-aligned memory, so it is the weakest guarantee those allocators share
+/// rather than the strongest any of them happens to provide.
+const MIN_ALIGN: usize = 2 * std::mem::size_of::<usize>();
```
daxhuiberts, line 32: This sound kinda guessy, there is probably some more fact based description possible here.

src/alloc.rs
```diff
@@ -40,16 +91,259 @@ unsafe impl GlobalAlloc for RedisAlloc {
...
+    #[test]
+    #[cfg_attr(miri, ignore = "cannot access the RedisModule_* extern statics")]
+    fn the_global_allocator_serves_over_aligned_layouts() {
+        #[repr(align(64))]
+        struct OverAligned([u8; 256]);
+
+        unsafe {
+            raw::RedisModule_Alloc = Some(zmalloc);
+            raw::RedisModule_Free = Some(zfree);
+        }
```
daxhuiberts, line 319: You're overriding global allocator functions here, but they point to a thread local. We're also not storing the old allocator functions back anymore at the end. The only way to test this function is to make sure no other tests are running in parallel.

The reason this looks like it works is the same reason the old allocator looked like it works. We're probably not hitting the big UB we're setting up here.

daxhuiberts (changes requested):
The fundamental implementation looks good, but lots of stuff around it needs improving.

This PR is part of redismodule-rs, but it is important to have an associated PR in RediSearch that uses this commit to test on RediSearch's CI as well, preferably an exhaustive CI run including benchmarks.

Because also for me this is dangerous stuff where dragons live and I'm not touching it lightly.

I'm also OK with taking it over if you prefer.

PR #485, opened 2026-09-21
Add hash_get_by_string: HGET with a RedisString field

GuyAv46:
## What

Adds `RedisKey::hash_get_by_string` and `RedisKeyWritable::hash_get_by_string`, plus the `raw::hash_get_by_string` helper: `HGET` with the field passed as an existing `RedisString`.

## Why

`hash_get` and `hash_get_multi` pass fields through `REDISMODULE_HASH_CFIELDS`. Per call that costs a `CString` and a result `Vec` on the Rust side, and inside Redis a `strlen` plus a transient string object (robj + sds) for the field name, created and freed around the lookup. For a module that fetches the same fields for many keys, that is five allocations and five frees per field that carry no information Redis does not already have.

With a `RedisModuleString` field and `REDISMODULE_HASH_NONE`, Redis reads the field in place and the only allocation left is the returned value. In RediSearch, switching document loading to this form cut server CPU per request by 16 to 21 percent on `FT.SEARCH ... RETURN` and `FT.AGGREGATE ... LOAD` of 32 fields (RediSearch/RediSearch#11509, MOD-18811).

## Notes

- Same ownership as `hash_get`: the returned value is bound to the key's context and freed through it.
- `None` for a missing field, and for a non-existent key on the read-only handle, matching `hash_get`.
- No behavior change to existing methods.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/raw.rs
```diff
@@ -586,6 +586,37 @@ where
...
 
+/// `HGET` with the field given as an existing `RedisModuleString`.
+///
+/// [`hash_get_multi`] passes fields as C strings (`REDISMODULE_HASH_CFIELDS`), which costs
+/// a `CString` here and a transient string object per field inside Redis. With a
+/// `RedisModuleString` Redis reads the field in place, so a caller that fetches the same
+/// field repeatedly can build the string once and reuse it.
+///
+/// Returns a null pointer when the hash has no such field. The returned value, when
+/// non-null, is a new string the caller must free.
```
oshadmi: Maybe a clearer comment?
```suggestion
/// Returns a null pointer when the hash has no such field. The returned value, when
/// a non-null result is an owned string reference that the caller must release
/// with `RedisModule_FreeString`.
```

src/key.rs
```diff
@@ -159,6 +159,22 @@ impl RedisKey {
         Ok(val)
     }
 
+    /// Like [`Self::hash_get`], with the field given as an existing [`RedisString`].
+    ///
+    /// Avoids the per-call `CString` and the transient field object Redis creates for
+    /// C-string fields; prefer it when the same field is fetched for many keys.
+    /// `None` when the key does not exist or has no such field.
+    pub fn hash_get_by_string(
```
oshadmi, line 167: Add integration coverage? (it introduces a separate FFI call that the existing tests don't cover)

src/raw.rs
GuyAv46: Applied in 701bc2a42, with a small wording tweak: "A non-null result is an owned string the caller must release with `RedisModule_FreeString`."

src/key.rs
GuyAv46: Added in 701bc2a42: a `map.hget` command in the `response` example module that goes through `hash_get_by_string`, and `test_response` now asserts the present-field, missing-field, and missing-key cases against a live Redis.

tests/integration.rs
```diff
@@ -577,6 +577,25 @@ fn test_response() -> Result<()> {
     res.sort();
     assert_eq!(&res, &["b", "d"]);
 
+    // hash_get_by_string: present field, absent field, absent key.
```
oshadmi, line 580: Nitpick: What about Wrong-type key? (and maybe even Writable handle (e.g., `map.hget_writable`))

GuyAv46: Both added in b5e0d7d11: a string key now fails with the "not a hash value" error on both handles, and a `map.hget_writable` command exercises `RedisKeyWritable::hash_get_by_string` for the present and missing field cases.

oshadmi (approved):

